PR #46, opened 2024-05-13
feat: NanoTDF Implementation

mustyantsev:


sdk/pom.xml
```diff
@@ -10,8 +10,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>11</source>
-                    <target>11</target>
+                    <source>15</source>
```
ttschampel, line 13: I think we want to keep this at 11 if possible until some downstream projects can upgrade (gateway)

ttschampel (commented):
Can we add a section in the README about using the Nano Encrypt feature?

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
```diff
@@ -79,4 +83,63 @@ public static Consumer<TDFConfig> withMetaData(String metaData) {
...
+        };
+    }
+
+    public static Consumer<NanoTDFConfig> withEllipticCurve(String curve) {
+        NanoTDFType.ECCurve ecCurve;
+        if (curve.compareToIgnoreCase(NanoTDFType.ECCurve.SECP384R1.toString()) == 0) {
+            ecCurve = NanoTDFType.ECCurve.SECP384R1;
+        } else if (curve.compareToIgnoreCase(NanoTDFType.ECCurve.SECP521R1.toString()) == 0) {
+            ecCurve = NanoTDFType.ECCurve.SECP521R1;
+        } else {
```
patmantru, line 136: should have at least a warning here that the supplied curve string was not recognized before using the default.

sdk/src/main/java/io/opentdf/platform/sdk/NanoTDF.java
```diff
@@ -0,0 +1,270 @@
...
+        return nanoTDFSize;
+    }
+
+    public void readNanoTDF(ByteBuffer nanoTDF, OutputStream outputStream,
+                            SDK.KAS ka) {
+
+
+
+
+        /*
```
patmantru, line 161: lol...looks familiar.  :)

sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ECKeyPair.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+    private KeyPair keyPair;
+    private String curveName;
+
+    public ECKeyPair() throws NoSuchAlgorithmException, NoSuchProviderException, InvalidAlgorithmParameterException {
+        this("secp256r1", ECAlgorithm.ECDH);
+    }
+
+    public ECKeyPair(String curveName, ECAlgorithm algorithm) throws NoSuchAlgorithmException, InvalidAlgorithmParameterException,
+            NoSuchProviderException {
```
patmantru, line 64: the signature says it throws those exceptions, but I don't see where they would be coming from.  It *should be* checking parms and throwing those as appropriate though.

sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/NanoTDFType.java
```diff
@@ -0,0 +1,43 @@
...
+public class NanoTDFType {
+    public enum ECCurve {
+        SECP256R1("secp256r1"),
+        SECP384R1("secp384r1"),
+        SECP521R1("secp384r1"),
+        SECP256K1("secp256k1");
+
+        private String name;
+
+        ECCurve(String curveName) {
```
patmantru, line 12: Should have argument validation here - throw exception if the string does not match one of the enum values

patmantru (changes requested):

patmantru (commented):
Changing to comment-only I think?

ttschampel:
Any idea why the checks aren't running for this PR?

sdk/pom.xml
ttschampel: The build is breaking b/c of this

sujankota: will fix it.

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
biscoe916: I agree. If no curve is provided, we should default, but if an invalid curve was supplied we should fail.

patmantru (dismissed):

patmantru (approved):

Issue #57, opened 2024-05-28
API calls with custom truststore


ttschampel:
Support additional truststore customization that is usable by GRPC comms to support self-signed certs or other untrusted CAs using https://github.com/Hakky54/sslcontext-kickstart + The NettyChannelBuilder

ttschampel:
completed

PR #58, opened 2024-05-28
feat(sdk): add ssl context

ttschampel:
- Issue #57 - Add kickstart SSLFactory to allow the configuration of a custom truststore for communications.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,7 @@
         <log4j.version>2.20.0</log4j.version>
         <grpc.version>1.63.0</grpc.version>
         <protobuf.version>3.25.3</protobuf.version>
+        <sslcontext.version>8.3.5</sslcontext.version>
     </properties>
     <modules>
         <module>protocol</module>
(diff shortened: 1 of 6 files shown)


.github/workflows/checks.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@5896cecc08fd8a1fbdfaf517e29b571164b031f7
         with:
-          java-version: "11"
+          java-version: "17"
```
mkleene, line 44: Gateway is currently at `11` so I think we need to live with the terrible strings for now.

pom.xml
```diff
@@ -12,11 +12,12 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 <!--        TODO: Move to new version of Java-->
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
         <log4j.version>2.20.0</log4j.version>
         <grpc.version>1.63.0</grpc.version>
         <protobuf.version>3.25.3</protobuf.version>
```
mkleene, line 19: nit, don't really care at all but maybe use the BOM here? https://mvnrepository.com/artifact/io.github.hakky54/sslcontext-kickstart-bom

sdk/src/main/java/io/opentdf/platform/sdk/SDKBuilder.java
```diff
@@ -135,12 +190,22 @@ public SDK build() {
     }
 
     private ManagedChannelBuilder<?> getManagedChannelBuilder() {
-        ManagedChannelBuilder<?> channelBuilder = ManagedChannelBuilder.forTarget(platformEndpoint);
-
-        if (usePlainText) {
-            channelBuilder = channelBuilder.usePlaintext();
```
mkleene, line 141: I screwed this up a bit by duplicating code. https://github.com/opentdf/java-sdk/pull/59 should make it so that there's only one place that we create channels.

mkleene (commented):
Is it possible to use something like this: https://stackoverflow.com/a/47638460/285915 to do this. I'd rather not pull in netty on the client side if we don't need to. OTOH it's probably not a big deal if there is a real reason.

.github/workflows/checks.yaml
ttschampel: rolled back to java 11

sdk/src/main/java/io/opentdf/platform/sdk/SDKBuilder.java
ttschampel: updated

ttschampel:
> Is it possible to use something like this: https://stackoverflow.com/a/47638460/285915 to do this. I'd rather not pull in netty on the client side if we don't need to. OTOH it's probably not a big deal if there is a real reason.

Updated to use approach you linked

pom.xml
```diff
@@ -69,6 +70,45 @@
...
+                        <artifactId>slf4j-api</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>io.grpc</groupId>
+                <artifactId>grpc-netty-shaded</artifactId>
+                <version>${grpc.version}</version>
+                <scope>runtime</scope>
+            </dependency>
```
mkleene, line 111:
```suggestion
```

sdk/pom.xml
```diff
@@ -111,5 +99,23 @@
...
+            <artifactId>sslcontext-kickstart-for-pem</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>sslcontext-kickstart</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>sslcontext-kickstart-for-netty</artifactId>
+        </dependency>
```
mkleene, line 113:
```suggestion
```

mkleene (approved):
Looks good! Tests are great. I _think_ that we can get rid of the dependencies I flagged but if that's not possible then fine.

sdk/pom.xml
ttschampel: Added Issue #62 to track this

pom.xml
ttschampel: Added Issue https://github.com/opentdf/java-sdk/issues/62 to track this

Issue #60, opened 2024-05-29
Expose SDK methods


ttschampel:
SDK methods/interfaces are currently scoped to default visibility, therefore consumers of the SDK can't access these.

PR #61, opened 2024-05-29
feat(sdk): Issue #60 - expose SDK

ttschampel:
Expose SDK for downstream use.


mkleene:
I'm fine with this now but I think that the final API we will want will be more like

```java
var sdk = SDKBuilder.build();
var tdfConfig = TDFConfig().withThis().withThat();
var tdf = sdk.createTDF(tdfConfig, data);

var payload = sdk.loadTDF(tdf);
```

I think it's fine to merge this but we might want to change it later when we have a slightly nicer API.

mkleene (approved):

PR #69, opened 2024-06-05
feat: add code coverage

mustyantsev:


mkleene (commented):
Do we want to have a minimum set so that we fail the build if it's not met?

ttschampel (commented):
ditto morgan's comment. we should add some rules to enforce on the coverage metrics

e.g were jacoco.line.coverage is a maven property with the min line coverage threshold for successful build
```
   <execution>
                            <id>check</id>
                            <goals>
                                <goal>check</goal>
                            </goals>
                            <configuration>
                                <rules>
                                    <rule>
                                        <element>BUNDLE</element>
                                        <limits>
                                            <limit>
                                                <counter>LINE</counter>
                                                <value>COVEREDRATIO</value>
                                                <minimum>${jacoco.line.coverage}</minimum>
                                            </limit>
                                        </limits>
                                    </rule>
                                </rules>
                            </configuration>
                        </execution>
```

Issue #71, opened 2024-06-07
Error using NanoTDF Encrypt


ttschampel:
Running the SDK against a deployment OpenTDF Platform to perform NanoTDF Encryption:

```
opentdf-nifi-1 |Caused by: java.lang.ClassCastException: class org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey cannot be cast to class org.bouncycastle.jce.interfaces.ECPublicKey (org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey is in unnamed module of loader org.apache.nifi.nar.NarClassLoader @2b2a4332; org.bouncycastle.jce.interfaces.ECPublicKey is in unnamed module of loader org.apache.nifi.nar.NarClassLoader @b01cb8d)
opentdf-nifi-1 |	at io.opentdf.platform.sdk.nanotdf.ECKeyPair.publicKeyFromPem(ECKeyPair.java:236)
opentdf-nifi-1 |	at io.opentdf.platform.sdk.NanoTDF.createNanoTDF(NanoTDF.java:77)
opentdf-nifi-1 |	at io.opentdf.nifi.ConvertToNanoTDF.lambda$processFlowFiles$0(ConvertToNanoTDF.java:69)
```

sujankota:
The issue looks like we are not getting the KAS public key.

ECPublicKey kasPublicKey = ECKeyPair.publicKeyFromPem(kasPublicKeyAsPem);

Probably check for empty string or check if it's not a PEM key throw correct exception.

ttschampel:
Fixed by #74

mkleene:
We've decided to go with a documentation-only approach (as opposed to loading the bouncy castle crypto provider ourselves) for now and see how that works for consumers.

PR #74, opened 2024-06-08
chore: Document bouncy castle usage

ttschampel:
Address downstream use wrt Bouncy Castle Provider registration: #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@
 
 OpenTDF Java SDK
 
+### SDK Usage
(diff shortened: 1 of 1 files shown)


mkleene (approved):

mkleene:
Do you think it's too much to do this ourselves if they try to do nano?

ttschampel:
@mkleene I think you are right, we should do that registration on Nano SDK usage

PR #82, opened 2024-06-23
fix(sdk): assertion support in tdf3

sujankota:


sdk/src/main/java/io/opentdf/platform/sdk/Assertion.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public enum StatementFormat {
+        ReferenceStatement("ReferenceStatement"),
+        StructuredStatement("StructuredStatement"),
+        StringStatement("StringStatement"),
+        Base64BinaryStatement("Base64BinaryStatement"),
+        XMLBase64("XMLBase64"),
+        HandlingStatement("HandlingStatement"),
+        StringType("String");
```
ttschampel, line 61: Maybe we should remove these for now since they don't match the examples in the spec (e.g. json-structured)  and until we have a use case to marshal/unmarshal these examples.

sdk/src/main/java/io/opentdf/platform/sdk/Assertion.java
```diff
@@ -0,0 +1,107 @@
...
+
+        private String format;
+
+        StatementFormat(String format) {
+            this.format = format;
+        }
+
+        @Override
+        public String toString() {
```
ttschampel, line 70: use built-in name function vs tostring() for comparison

ttschampel (approved):
Adding a ticket #83 to add E2E tests for this feature.

mustyantsev (approved):

PR #106, opened 2024-08-02
feat(core): Identifier in NanoTDF ResourceLocator borrowed from Protocol

pflynn-virtru:
Add support for identifiers in ResourceLocator

Introduced an IdentifierType enum to categorize different identifier lengths. Updated ResourceLocator constructors to handle identifiers, including parsing and setting appropriate identifier types. Modified ByteBuffer constructor to extract identifier information from the protocol byte.

Issue: https://github.com/opentdf/platform/issues/1203
Specification: https://github.com/opentdf/spec/pull/40
ADR: https://github.com/opentdf/platform/issues/900


sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ResourceLocator.java
```diff
@@ -73,17 +147,32 @@ public int writeIntoBuffer(ByteBuffer buffer) {
...
         totalBytesWritten += 1;
 
-        // Write the url body;
+        // Write the url body
         buffer.put(body);
         totalBytesWritten += body.length;
 
+        // Write the identifier
+        if (identifierType != NanoTDFType.IdentifierType.NONE) {
+            buffer.put(identifier);
```
dmihalcik-virtru, line 168: You may need to pad the buffer here or assert (throw IllegalArg) if identifier is. the wrong length

pflynn-virtru: Note we have this setter (without getter)
```
public void setProtocol()
```

pflynn-virtru: done added padded and tests

sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ResourceLocator.java
```diff
@@ -1,37 +1,140 @@
...
+                throw new IllegalArgumentException("Unexpected identifier type: " + identifierType);
+        }
+    }
+
+    public void setIdentifier(String identifier) {
+        if (identifier == null) {
+            this.identifierType = NanoTDFType.IdentifierType.NONE;
+            this.identifier = new byte[0];
+        } else {
+            byte[] identifierBytes = identifier.getBytes();
```
mkleene, line 116: I think we should set the charset here. If we [don't it uses the platform default](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes--)

mkleene (commented):
Looks good except for the one comment!

dmihalcik-virtru:
merged as https://github.com/opentdf/java-sdk/pull/112

PR #108, opened 2024-08-09
feat(sdk): add mime type.

c-r33d:
- Add withMimeType TDF Config option, which will in-turn add it to the Manifest.Payload.MimeType
- Add getManifest() function to the reader object. In-sync with Golang SDK.

### DIFF
--- a/sdk/src/main/java/io/opentdf/platform/sdk/Config.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/Config.java
@@ -16,6 +16,7 @@ public class Config {
     public static final int TDF3_KEY_SIZE = 2048;
     public static final int DEFAULT_SEGMENT_SIZE = 2 * 1024 * 1024; // 2mb
     public static final String KAS_PUBLIC_KEY_PATH = "/kas_public_key";
+    public static final String DEFAULT_MIME_TYPE = "application/octet-stream";
 
     public enum TDFFormat {
         JSONFormat,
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/io/opentdf/platform/sdk/Config.java
```diff
@@ -16,6 +16,7 @@ public class Config {
     public static final int TDF3_KEY_SIZE = 2048;
     public static final int DEFAULT_SEGMENT_SIZE = 2 * 1024 * 1024; // 2mb
     public static final String KAS_PUBLIC_KEY_PATH = "/kas_public_key";
+    public static final String DEFAULT_MIME_TYPE = "application/octet-stream";
```
dmihalcik-virtru, line 19: I think the other clients default to no mime type (the field is dropped)

dmihalcik-virtru (approved):

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
jrschumacher: CLI defaults to `application/octet-stream` which was modeled from MDM web docs. We can change this if needed. https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

> application/octet-stream is the default value for all other cases. An unknown file type should use this type. Browsers are particularly careful when manipulating these files to protect users from software vulnerabilities and possible dangerous behavior.

c-r33d: https://github.com/opentdf/platform/blob/main/sdk/tdf.go#L252

jrschumacher (approved):

PR #109, opened 2024-08-13
feat(core): Handle split keys on tdf3 encrypt and decrypt

elizabethhealy:
equivalent to https://github.com/opentdf/platform/pull/1038


sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -495,22 +504,46 @@ private void fillInPublicKeyInfo(List<Config.KASInfo> kasInfoList, SDK.KAS kas)
     }
 
     public Reader loadTDF(SeekableByteChannel tdf, Config.AssertionConfig assertionConfig, SDK.KAS kas) throws NotValidateRootSignature, SegmentSizeMismatch,
-            IOException, FailedToCreateGMAC, JOSEException, ParseException, NoSuchAlgorithmException, DecoderException {
+            IOException, FailedToCreateGMAC, JOSEException, ParseException, NoSuchAlgorithmException, DecoderException, SplitKeyException {
```
dmihalcik-virtru, line 507: I'd recommend against adding a new checked `Exception` here since it will count as a semver breaking change. Instead extend an existing checked exception or let `SplitKeyException` extend (hopefully some more specific subclass of) `RuntimeException`

dmihalcik-virtru (changes requested):

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -495,22 +504,46 @@ private void fillInPublicKeyInfo(List<Config.KASInfo> kasInfoList, SDK.KAS kas)
...
 
         TDFReader tdfReader = new TDFReader(tdf);
         String manifestJson = tdfReader.manifest();
         Manifest manifest = gson.fromJson(manifestJson, Manifest.class);
         byte[] payloadKey = new byte[GCM_KEY_SIZE];
         String unencryptedMetadata = null;
+        
+        Map<String, Boolean> knownSplits = new HashMap<>();
+        Map<String, Boolean> foundSplits = new HashMap<>();
```
mkleene, line 516: could these be `Set`s?

mkleene (commented):
A few minor suggestions that aren't important but could we also add some tests?

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -495,22 +504,46 @@ private void fillInPublicKeyInfo(List<Config.KASInfo> kasInfoList, SDK.KAS kas)
...
+          !manifest.encryptionInformation.keyAccessObj.get(0).sid.isEmpty();
 
         MessageDigest digest =  MessageDigest.getInstance("SHA-256");
 
         if (manifest.payload.isEncrypted) {
             for (Manifest.KeyAccess keyAccess: manifest.encryptionInformation.keyAccessObj) {
-                var unwrappedKey = kas.unwrap(keyAccess, manifest.encryptionInformation.policy);
+                Config.SplitStep ss = new Config.SplitStep(keyAccess.url, keyAccess.sid);
+                byte[] unwrappedKey;
+                if (!mixedSplits) {
```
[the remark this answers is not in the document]
elizabethhealy: if we negate the variable it becomes notMixedSplits which becomes confusing logic wise and there is a check for if(mixedsplits) later in the code so we would end up with `if(!notMixedSplits)` which is kind of confusing

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -164,57 +171,111 @@ PolicyObject createPolicyObject(List<String> attributes) {
...
+                for (Config.KASInfo kasInfo : tdfConfig.kasInfoList) {
+                    Config.SplitStep step = new Config.SplitStep(kasInfo.URL, "");
+                    if (tdfConfig.kasInfoList.size() > 1) {
+                        step.splitID = String.format("s-%d", splitPlan.size());
+                    }
+                    splitPlan.add(step);
+                    if (kasInfo.PublicKey != null && !kasInfo.PublicKey.isEmpty()) {
+                        latestKASInfo.put(kasInfo.URL, kasInfo);
+                    }
+                }
```
dmihalcik-virtru, line 194: I think the other clients create a split, not shared key by default. you'll have to add a counter 

```suggestion
                int i = 0;
                for (Config.KASInfo kasInfo : tdfConfig.kasInfoList) {
                    Config.SplitStep step = new Config.SplitStep(kasInfo.URL, "");
                    if (tdfConfig.kasInfoList.size() > 1) {
                        step.splitID = String.format("s-%d", i++);
                    }
                    splitPlan.add(step);
                    if (kasInfo.PublicKey != null && !kasInfo.PublicKey.isEmpty()) {
                        latestKASInfo.put(kasInfo.URL, kasInfo);
                    }
                }
```

dmihalcik-virtru (changes requested):

dmihalcik-virtru (approved):

PR #111, opened 2024-08-13
feat: BACK-2316 add a simple method to detect TDFs

mkleene:
* just validate that we have a zip file with the right entries
* add `InvalidZipException` so that we can catch it


sdk/src/main/java/io/opentdf/platform/sdk/SDK.java
```diff
@@ -112,4 +118,25 @@ public Optional<ClientInterceptor> getAuthInterceptor() {
...
         return this.services;
     }
+
+    /**
+     * Checks to see if this has the structure of a Z-TDF in that it is a zip file containing
+     * a `manifest.json` and a `0.payload`
+     * @param channel A channel containing the bytes of the potential Z-TDF
+     * @return `true` if
+     */
+    public static boolean mightBeZTDF(SeekableByteChannel channel) {
```
strantalis, line 128: @mkleene 

Can we name this similar to what we did in the go sdk. `IsValidTdf`. In the future we can do the same json schema validation too.

biscoe916: Can we go with isTDF, and isValidTdf? I'd also be open to a single function that takes two arguments, with the second argument saying whether or not we want to do a deeper validation check.

@mkleene @strantalis Thoughts?

strantalis (approved):

PR #117, opened 2024-08-15
feat(sdk): Update the assertion support to match go sdk

sujankota:


strantalis (approved):

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
```diff
@@ -36,21 +34,41 @@ public static class KASInfo {
...
+        }
+    }
+
+    public static class TDFReaderConfig {
+        // Optional Map of Assertion Verification Keys
+        AssertionVerificationKeys assertionVerificationKeys;
+    }
+
+    @SafeVarargs
+    public static TDFReaderConfig newTTDFReaderConfig(Consumer<TDFReaderConfig>... options) {
```
mkleene, line 62: maybe `newTDFReaderConfig`

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
```diff
@@ -36,21 +34,41 @@ public static class KASInfo {
...
+        TDFReaderConfig config = new TDFReaderConfig();
+        for (Consumer<TDFReaderConfig> option : options) {
+            option.accept(config);
         }
+        return config;
+    }
+
+    public static Consumer<TDFReaderConfig> WithAssertionVerificationKeys(AssertionVerificationKeys assertionVerificationKeys) {
```
mkleene, line 70: I don't think that we should capitalize this.

sdk/src/main/java/io/opentdf/platform/sdk/AssertionConfig.java
```diff
@@ -0,0 +1,103 @@
...
+            return Objects.equals(format, statement.format) && Objects.equals(schema, statement.schema) && Objects.equals(value, statement.value);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(format, schema, value);
+        }
+    }
+
+    public String id;
```
mkleene, line 97: Do you think that it would be clearer if this were a `keyId`? That naming seems like it might be easier to understand but I'm not sure if it's a good idea since we'd need to get into all of the other stuff with key ids. If we keep doing it this way, though, I think that we will have our own set of problems with tracking assertion ids.

sdk/src/main/java/io/opentdf/platform/sdk/Manifest.java
```diff
@@ -215,6 +230,170 @@ public int hashCode() {
...
+            return Objects.equals(method, binding.method) && Objects.equals(signature, binding.signature);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(method, signature);
+        }
+    }
+
+    static public class Assertion {
```
mkleene, line 251: I like the separation of the `AssertionConfig` and `Assertion` classes but I think we might be able to get almost the same thing by making the `sign`, `verify`, etc. methods private. I think that _maybe_ they should be unless we want to expose them to clients of the SDK. Not sure about that at all.

mkleene (approved):
Looks good, it's definitely an improvement. I think it's fine to merge as is but I had a couple of questions and some potential renames if you feel like it!

mkleene (approved):

mkleene (approved):

Issue #103, opened 2024-07-29
feat: attribute based key splits


strantalis:
With the introduction of attribute based key splitting in the go sdk. We now need to introduce this for ztdf in the java-sdk. Implement the key splitting logic that was done in the go sdk and update x-tests to make
key splits generated with each sdk can interop.

PR #120, opened 2024-08-19
feat(core): Add autoconfigure for key splitting

elizabethhealy:
resolves https://github.com/opentdf/java-sdk/issues/103


.github/workflows/checks.yaml
```diff
@@ -139,7 +139,7 @@ jobs:
             --client-secret=secret \
             --platform-endpoint=localhost:8080 \
             -i \
-            encrypt --kas-url=localhost:8080 --mime-type=text/plain --attr https://example.com/attr/attr1/value/value1 -f data -m 'here is some metadata' > test.tdf
+            encrypt --kas-url=localhost:8080 --mime-type=text/plain --attr https://example.com/attr/attr1/value/value1 --autoconfigure=false -f data -m 'here is some metadata' > test.tdf
```
elizabethhealy, line 142: i dont have a local e2e test for autoconfigure yet

elizabethhealy:
successful xtest run https://github.com/opentdf/tests/actions/runs/10473128094/job/29004376734?pr=166

cmdline/src/main/java/io/opentdf/platform/Command.java
```diff
@@ -68,15 +66,17 @@ void encrypt(
...
-        });
-
+        if (attributes.isPresent()){
+            configs.add(Config.withDataAttributes(attributes.get().split(",")));
+        }
         var tdfConfig = Config.newTDFConfig(configs.toArray(Consumer[]::new));
         try (var in = file.isEmpty() ? new BufferedInputStream(System.in) : new FileInputStream(file.get())) {
             try (var out = new BufferedOutputStream(System.out)) {
-                new TDF().createTDF(in, out, tdfConfig, sdk.getServices().kas());
+                new TDF().createTDF(in, out, tdfConfig,
```
dmihalcik-virtru, line 77: Is this formatted? I feel like this needs some more indenting

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+import io.opentdf.platform.policy.attributes.GetAttributeValuesByFqnsRequest;
+import io.opentdf.platform.policy.attributes.GetAttributeValuesByFqnsResponse;
+import io.opentdf.platform.policy.attributes.GetAttributeValuesByFqnsResponse.AttributeAndValue;
+import io.opentdf.platform.policy.Attribute;
+import io.opentdf.platform.policy.Value;
+import io.opentdf.platform.policy.AttributeValueSelector;
+import io.opentdf.platform.policy.AttributeRuleTypeEnum;
+
+// Error handling class
+class AutoConfigureException extends IOException {
```
dmihalcik-virtru, line 35: This should be public, since it is on public signatures. Also I am not sure that IOException is quite right; it often indicates a configuration error on the backend, not a failure of network or file system, right?

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+    public AutoConfigureException(String message) {
+        super(message);
+    }
+}
+
+// Attribute rule types: operators!
+class RuleTypes {
```
dmihalcik-virtru, line 42: `RuleType` (singular) right?

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+    public static final String ALL_OF = "allOf";
+    public static final String ANY_OF = "anyOf";
+    public static final String UNSPECIFIED = "unspecified";
+    public static final String EMPTY_TERM = "DEFAULT";
+}
+
+
+public class Autoconfigure {
+
+    public static Logger logger = LoggerFactory.getLogger(TDF.class);
```
dmihalcik-virtru, line 53: Not `Autoconfigure.class`?

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+    public static class SplitStep {
+        public String kas;
+        public String splitID;
+
+        public SplitStep(String kas, String splitId) {
+            this.kas = kas;
+            this.splitID = splitId;
+        }
+
+        public boolean equals(Object obj) {
```
dmihalcik-virtru, line 64: If you define `equals` you should generally define `hashCode` so the value objects can be used in `HashSet` and as keys

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+import io.opentdf.platform.policy.attributes.GetAttributeValuesByFqnsResponse;
+import io.opentdf.platform.policy.attributes.GetAttributeValuesByFqnsResponse.AttributeAndValue;
+import io.opentdf.platform.policy.Attribute;
+import io.opentdf.platform.policy.Value;
+import io.opentdf.platform.policy.AttributeValueSelector;
+import io.opentdf.platform.policy.AttributeRuleTypeEnum;
+
+// Error handling class
+class AutoConfigureException extends IOException {
+    public AutoConfigureException(String message) {
```
dmihalcik-virtru, line 36: I recommend adding a constructor with a `cause` parameter, also, since that helps with debugging

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        
+        public boolean equals(Object obj) {
+            if (this == obj) {
+                return true;
+            }
+            if (obj == null || !(obj instanceof AttributeValueFQN)) {
+                return false;
+            }
+            AttributeValueFQN afqn = (AttributeValueFQN) obj;
+            if ((this.url.equals(afqn.url)) && (this.key.equals(afqn.key))){
```
dmihalcik-virtru, line 178: should this just compare on `key`? Also, see note above about `hashCode`

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            try {
+                return URLDecoder.decode(matcher.group(1), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
+                throw new AutoConfigureException("invalid type");
+            }
+        }
+    }
+
+    // Utility class for an attribute value FQN.
+    public static class AttributeValueFQN {
```
dmihalcik-virtru, line 143: You may want to make these public static classes top level classes (in new files)

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            this.policy = policy;
+        }
+
+        public Granter(List<AttributeValueFQN> policy, Map<String, KeyAccessGrant> grants) {
+            this.policy = policy;
+            this.grants = grants;
+        }
+
+        public  Map<String, KeyAccessGrant> getGrants() {
+            return grants;
```
dmihalcik-virtru, line 258: I'd suggest adding an immutable facade somewhere to avoid making this updatable, and/or copying it in the constructor

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        }
+
+        public List<AttributeValueFQN> getPolicy() {
+            return policy;
+        }
+
+        public void addGrant(AttributeValueFQN fqn, String kas, Attribute attr) {
+            grants.computeIfAbsent(fqn.key, k -> new KeyAccessGrant(attr, new ArrayList<>()))
```
dmihalcik-virtru, line 266: Do we not have a [guava](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ArrayListMultimap.html) dep? I don't think it is worth adding for this

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            private Attribute def;
+            private List<AttributeValueFQN> values;
+        
+            public SingleAttributeClause(Attribute def, List<AttributeValueFQN> values) {
+                this.def = def;
+                this.values = values;
+            }
+        }
+
+        public class AttributeBooleanExpression {
```
dmihalcik-virtru, line 432: These can be package protected
```suggestion
        class AttributeBooleanExpression {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                Attribute ad = this.dict.get(prefix);
+                if (ad == null) {
+                    throw new Exception("Unknown attribute type: [" + prefix.toString() + "], not in [" + this.dict.keySet().toString() + "]");
+                }
+                return ad;
+            }
+
+        }
+
+        public class SingleAttributeClause {
```
dmihalcik-virtru, line 421:
```suggestion
        static class SingleAttributeClause {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+
+        public String getKey() {
+            return key;
+        }
+
+        public String authority() throws AutoConfigureException {
+            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+)/attr/\\S*/value/\\S*$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find()) {
+                throw new AutoConfigureException("invalid type");
```
dmihalcik-virtru, line 192: This regex is validated in the constructor so it can't throw here.

```suggestion
        public String authority() {
            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+)/attr/\\S*/value/\\S*$");
            Matcher matcher = pattern.matcher(url);
            if (!matcher.find()) {
                throw new RuntimeException("invalid attribute");
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            }
+            return matcher.group(1);
+        }
+
+        public AttributeNameFQN prefix() throws AutoConfigureException {
+            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+/attr/\\S*)/value/\\S*$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find()) {
+                throw new AutoConfigureException("invalid type");
+            }
```
dmihalcik-virtru, line 202:
```suggestion
        public AttributeNameFQN prefix() {
            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+/attr/\\S*)/value/\\S*$");
            Matcher matcher = pattern.matcher(url);
            if (!matcher.find()) {
                throw new RuntimeException("invalid type");
            }
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        public String value() throws AutoConfigureException {
+            Pattern pattern = Pattern.compile("^https?://[\\w./-]+/attr/\\S*/value/(\\S*)$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find()) {
+                throw new AutoConfigureException("invalid type");
+            }
+            try {
+                return URLDecoder.decode(matcher.group(1), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
+                throw new AutoConfigureException("invalid type");
```
dmihalcik-virtru, line 215:
```suggestion
        public String value() {
            Pattern pattern = Pattern.compile("^https?://[\\w./-]+/attr/\\S*/value/(\\S*)$");
            Matcher matcher = pattern.matcher(url);
            if (!matcher.find()) {
                throw new AutoConfigureException("invalid type");
            }
            try {
                return URLDecoder.decode(matcher.group(1), StandardCharsets.UTF_8.name());
            } catch (UnsupportedEncodingException | IllegalArgumentException e) {
                throw new RuntimeException("invalid type", e);
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+
+        public String name() throws AutoConfigureException {
+            Pattern pattern = Pattern.compile("^https?://[\\w./-]+/attr/(\\S*)/value/\\S*$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find()) {
+                throw new AutoConfigureException("invalid attributeInstance");
+            }
+            try {
+                return URLDecoder.decode(matcher.group(1), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
```
dmihalcik-virtru, line 227:
```suggestion
            } catch (UnsupportedEncodingException | IllegalArgumentException e) {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                throw new AutoConfigureException("invalid attributeInstance");
```
dmihalcik-virtru, line 228:
```suggestion
                throw new RuntimeException("invalid attributeInstance");
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                throw new AutoConfigureException("invalid type");
+            }
+            try {
+                return URLDecoder.decode(matcher.group(1), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
+                throw new AutoConfigureException("invalid type");
+            }
+        }
+
+        public String name() throws AutoConfigureException {
```
dmihalcik-virtru, line 219:
```suggestion
        public String name() {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                throw new AutoConfigureException("invalid type");
+            }
+            return matcher.group(1);
+        }
+
+        public AttributeNameFQN prefix() throws AutoConfigureException {
+            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+/attr/\\S*)/value/\\S*$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find()) {
+                throw new AutoConfigureException("invalid type");
```
dmihalcik-virtru, line 201:
```suggestion
        public AttributeNameFQN prefix() {
            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+/attr/\\S*)/value/\\S*$");
            Matcher matcher = pattern.matcher(url);
            if (!matcher.find()) {
                throw new RuntimeException("invalid type");
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            Pattern pattern = Pattern.compile("^(https?://[\\w./-]+)/attr/(\\S*)/value/(\\S*)$");
+            Matcher matcher = pattern.matcher(url);
+            if (!matcher.find() || matcher.group(1) == null || matcher.group(2) == null || matcher.group(3) == null) {
+                throw new AutoConfigureException("invalid type: attribute regex fail for [" + url + "]");
+            }
+
+            try {
+                URLDecoder.decode(matcher.group(2), StandardCharsets.UTF_8.name());
+                URLDecoder.decode(matcher.group(3), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
```
dmihalcik-virtru, line 157:
```suggestion
            } catch (UnsupportedEncodingException | IllegalArgumentException e) {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        public KeyAccessGrant(Attribute attr, List<String> kases) {
+            this.attr = attr;
+            this.kases = kases;
+        }
+    }
+
+    // Structure capable of generating a split plan from a given set of data tags.
+    public static class Granter {
+        private final List<AttributeValueFQN> policy;
+        private Map<String, KeyAccessGrant> grants = new HashMap<>();
```
dmihalcik-virtru, line 246:
```suggestion
        private final Map<String, KeyAccessGrant> grants = new HashMap<>();
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        }
+
+        public void addGrant(AttributeValueFQN fqn, String kas, Attribute attr) {
+            grants.computeIfAbsent(fqn.key, k -> new KeyAccessGrant(attr, new ArrayList<>()))
+                .kases.add(kas);
+        }
+
+        public void addAllGrants(AttributeValueFQN fqn, List<KeyAccessServer> gs, Attribute attr) {
+            if (gs.isEmpty()) {
+                grants.putIfAbsent(fqn.key, new KeyAccessGrant(attr, new ArrayList<>()));
```
dmihalcik-virtru, line 272: if you implement hashCode and equals on `AttributeValueFQN` on just `key`, you could use the fqn as the key instead of the string

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+
+        public KeyAccessGrant byAttribute(AttributeValueFQN fqn) {
+            return grants.get(fqn.key);
+        }
+
+        public interface StringOperator {
+            public String op();
+        }
+
+
```
dmihalcik-virtru, line 290: Is this referenced somewhere?

```suggestion
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                for (int i = 0; i < values.size(); i++) {
+                    if (i > 0) {
+                        sb.append("&");
+                    }
+                    sb.append(values.get(i).toString());
+                }
+                return sb.toString();
+            }
+
+            public int len() {
```
dmihalcik-virtru, line 552: in java you generally say `size`

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                if (this.dict == null) {
+                    this.dict = new HashMap<>();
+                }
+        
+                AttributeNameFQN prefix;
+                try {
+                    prefix = new AttributeNameFQN(ad.getFqn());
+                } catch (Exception e) {
+                    throw new Exception(e);
+                }
```
dmihalcik-virtru, line 402: an AutoConfigureException here indicates an invalid attribute FQN in the database

```suggestion
            public void put(Attribute ad) throws AutoConfigureException {
                if (this.dict == null) {
                    this.dict = new HashMap<>();
                }
        
                AttributeNameFQN prefix;
                try {
                    prefix = new AttributeNameFQN(ad.getFqn());
                } catch (AutoConfigureException e) {
                    throw new AutoConfigureException(e);
                }
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+        
+                if (this.dict.containsKey(prefix)) {
+                    throw new Exception("Attribute prefix already found: [" + prefix.toString() + "]");
```
dmihalcik-virtru, line 405:
```suggestion
                    throw new AutoConfigureException("Attribute prefix already found: [" + prefix.toString() + "]");
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+                }
+        
+                this.dict.put(prefix, ad);
+            }
+        
+            public Attribute get(AttributeNameFQN prefix) throws Exception {
+                Attribute ad = this.dict.get(prefix);
+                if (ad == null) {
+                    throw new Exception("Unknown attribute type: [" + prefix.toString() + "], not in [" + this.dict.keySet().toString() + "]");
```
dmihalcik-virtru, line 414:
```suggestion
                    throw new AutoConfigureException("Unknown attribute type: [" + prefix.toString() + "], not in [" + this.dict.keySet().toString() + "]");
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            List<SingleAttributeClause> must = sortedPrefixes.stream()
+                    .map(prefixes::get)
+                    .collect(Collectors.toList());
+
+            return new AttributeBooleanExpression(must);
+        }
+
+
+
+        public class AttributeMapping {
```
dmihalcik-virtru, line 384:
```suggestion
        static class AttributeMapping {
```

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,749 @@
...
+            return new BooleanKeyExpression(kcs);
+        }
+
+        AttributeBooleanExpression constructAttributeBoolean() throws AutoConfigureException {
+            Map<String, SingleAttributeClause> prefixes = new HashMap<>();
+            List<String> sortedPrefixes = new ArrayList<>();
+            for (AttributeValueFQN aP : policy) {
+                AttributeNameFQN a = aP.prefix();
+                String p = a.toString().toLowerCase();
+                SingleAttributeClause clause = prefixes.get(p);
```
dmihalcik-virtru, line 365:
```suggestion
                SingleAttributeClause clause = prefixes.get(a.key);
```

dmihalcik-virtru (commented):
I think these are all java nits? Substantively, this is good. I haven't made it through the whole thing to review yet though so I thought I'd leave some notes. Mainly, I do think it is a bad idea to catch or throw `Exception`, and there are a lot of checked exceptions here that will never throw in practice (mostly since the FQN classes validate in their constructor, so you can't get the regex, url decode, etc. exceptions in the getters - in fact you may want to store the decoded strings instead of the encoded ones for performance if you expect to call those methods ~ 1x per object or more)

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
elizabethhealy: ya i was trying to extend an exception that we already throw from encrypt in command.java so there wouldnt be breaking changes but i agree it doesnt really make sense as the base exception

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
elizabethhealy: it still needs throws in the func definition bc after the matcher it create a new AttributeNameFqn and the constructor of that throws AutoConfigureException

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
elizabethhealy: dupe of above, see comment

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
elizabethhealy: returning a copy of the hashmap instead here

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
```diff
@@ -0,0 +1,741 @@
...
+
+            Attribute def = pair.getAttribute();
+            if (def != null) {
+                grants.addAllGrants(fqn, def.getGrantsList(), def);
+            }
+
+            Value v = pair.getValue();
+            if (v != null) {
+                grants.addAllGrants(fqn, v.getGrantsList(), def);
+            }
```
dmihalcik-virtru, line 706: We no longer add both grants from Value and Attribute. Instead, grants from Values, if they exist, overload any grants on attributes.

```suggestion
            Value v = pair.getValue();
            if (v != null && !v.getGrantsList().isEmpty()) {
                grants.addAllGrants(fqn, v.getGrantsList(), def);
            } else {
              Attribute def = pair.getAttribute();
              if (def != null) {
                  grants.addAllGrants(fqn, def.getGrantsList(), def);
              }
            }

```

dmihalcik-virtru (changes requested):

sdk/src/main/java/io/opentdf/platform/sdk/Autoconfigure.java
elizabethhealy: will do this as a follow on with the test from https://github.com/opentdf/platform/pull/1318/files

dmihalcik-virtru (approved):

PR #128, opened 2024-08-23
fix: ztdf support both base and handling assertions

sujankota:


ttschampel (commented):
Can we add a unit test to cover this change?

sujankota:
https://github.com/opentdf/java-sdk/blob/main/sdk/src/test/java/io/opentdf/platform/sdk/TDFTest.java#L210 - We have test for handling assertion

ttschampel:
That test (or another test) doesn't actually check that both a base assertion and handling assertion result in assertions in the output TDF as far as I can tell

sujankota:
Please check the updated test.

ttschampel (approved):

PR #137, opened 2024-08-26
chore(ci): Slow release please bumps

dmihalcik-virtru:
prefer patch for all but breaking changes, basically


dmihalcik-virtru:
added with https://github.com/opentdf/java-sdk/pull/139

PR #148, opened 2024-08-27
fix: Align identifier bytes correctly in ResourceLocator

pflynn-virtru:
Adjust the starting index of System.arraycopy to correctly align the identifier bytes within the given array length. This change ensures that shorter identifiers are aligned at the end of the array, improving data consistency and correctness.

Pad right

### DIFF
--- a/sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ResourceLocator.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ResourceLocator.java
@@ -60,15 +60,15 @@ public class ResourceLocator {
             } else if (identifierLen <= 2) {
                 this.identifierType = NanoTDFType.IdentifierType.TWO_BYTES;
                 this.identifier = new byte[NanoTDFType.IdentifierType.TWO_BYTES.getLength()];
-                System.arraycopy(identifier.getBytes(), 0, this.identifier, 0, identifierLen);
+                System.arraycopy(identifier.getBytes(), 0, this.identifier, this.identifier.length - identifierLen, identifierLen);
             } else if (identifierLen <= 8) {
                 this.identifierType = NanoTDFType.IdentifierType.EIGHT_BYTES;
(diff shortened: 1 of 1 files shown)


dmihalcik-virtru (changes requested):
Add some golden content tests, similar to the ones here: https://github.com/opentdf/client-web/pull/338/files#diff-bf61ddc3e1b6824287643ae59fd2d22f3305e3495e9a8380576d671492ac769e

I'm pretty sure the current code is correct for encoding, but yes you should add the trim method on line 224 (I think? I'm not sure that new string doesn't do this for you)

pflynn-virtru:
> Add some golden content tests, similar to the ones here: https://github.com/opentdf/client-web/pull/338/files#diff-bf61ddc3e1b6824287643ae59fd2d22f3305e3495e9a8380576d671492ac769e
> 
> I'm pretty sure the current code is correct for encoding, but yes you should add the trim method on line 224 (I think? I'm not sure that new string doesn't do this for you)

- added golden identifier tests
- getIdentifier() will return byte[] without trim, getIdentifierString() does trim

sdk/src/main/java/io/opentdf/platform/sdk/nanotdf/ResourceLocator.java
```diff
@@ -183,13 +183,21 @@ public int writeIntoBuffer(ByteBuffer buffer) {
...
+                buffer.put((byte) protocol.ordinal());
+                break;
+            case TWO_BYTES:
+                buffer.put((byte) ((0b0001 << 4) | protocol.ordinal()));
+                break;
+            case EIGHT_BYTES:
+                buffer.put((byte) ((0b0010 << 4) | protocol.ordinal()));
+                break;
+            case THIRTY_TWO_BYTES:
+                buffer.put((byte) ((0b0100 << 4) | protocol.ordinal()));
```
dmihalcik-virtru, line 197: Isn't this 3?
```suggestion
                buffer.put((byte) ((0b0011 << 4) | protocol.ordinal()));
```

dmihalcik-virtru (approved):

PR #154, opened 2024-08-30
feat(cli): Add insecure option for https connections

elizabethhealy:
Currently --insecure forces http plaintext. Instead have --insecure use an ssl factory that does not check certs. Also add a flag for http plaintext (-h).

### DIFF
--- a/cmdline/src/main/java/io/opentdf/platform/Command.java
+++ b/cmdline/src/main/java/io/opentdf/platform/Command.java
@@ -93,7 +93,6 @@ class Command {
     }
 
     private SDK buildSDK() {
-        // Create an SSLFactory with the TrustManager that trusts all certificates
         SDKBuilder builder = new SDKBuilder();
         if (insecure){
             SSLFactory sslFactory = SSLFactory.builder()
--- a/cmdline/src/main/java/io/opentdf/platform/Command.java
+++ b/cmdline/src/main/java/io/opentdf/platform/Command.java
@@ -42,7 +42,7 @@ class Command {
     @Option(names = {"--client-secret"}, required = true)
     private String clientSecret;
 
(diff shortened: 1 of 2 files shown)


elizabethhealy:
xtest failing due to flag changes but i have a PR to fix it after this merges https://github.com/opentdf/tests/pull/185

cmdline/src/main/java/io/opentdf/platform/Command.java
```diff
@@ -84,10 +93,17 @@ void encrypt(
...
+            SSLFactory sslFactory = SSLFactory.builder()
+            .withUnsafeTrustMaterial() // Trust all certificates
+            .build();
+            builder.sslFactory(sslFactory);
+        }
+        
+        return  builder.platformEndpoint(platformEndpoint)
                 .clientSecret(clientId, clientSecret)
-                .useInsecurePlaintextConnection(insecure)
+                .useInsecurePlaintextConnection(plaintext)
```
strantalis, line 106: Will this ever conflict with the sslFactory?

elizabethhealy: sslfactory is only used if plaintext isnt set, when plaintext set it just defaults to http and doesnt use the factory
https://github.com/opentdf/java-sdk/blob/7069b1ba5ddc7098a7997cdcfcfdaddb6f899b63/sdk/src/main/java/io/opentdf/platform/sdk/SDKBuilder.java#L227-L237

elizabethhealy: if neither are set it looks like it just does `channelBuilder = ManagedChannelBuilder.forTarget(endpoint);`

strantalis (approved):

Issue #156, opened 2024-09-01
move to single module


strantalis:
Right now the java-sdk consists of multiple modules. There is really no reason for us to do this as it increases our complexity when it comes to releases.

Move the java-sdk to a single module and update any release-please configuration to support this.

Acceptance Criteria:
- Single jar is published for protocol and sdk
- Documentation updated

PR #160, opened 2024-09-10
refactor!: move to single jar

strantalis:
resolves: https://github.com/opentdf/java-sdk/issues/156


mkleene (commented):
I'd rather keep the command line around and publish a combined sdk/protocol. Unless there are use cases I don't know about the clients of the command line are CI and the clients of the SDK are applications that don't care about the command line. Forcing them to pull in a bunch of code they don't need doesn't seem ideal to me.

mkleene (commented):
I don't think that we even need to publish the command line

strantalis:
@mkleene Will the following ignore the cli deps and will maven actually publish the cli jar if its not the primary ?

https://github.com/opentdf/java-sdk/pull/160/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R310-R319

mkleene (approved):

PR #163, opened 2024-09-17
fix(sdk): Mixed split fix

dmihalcik-virtru:


mkleene (approved):
It would be better to have a test but it _looks_ like xtest should catch this.

elizabethhealy (approved):
ooh ya this is v wrong, good catch ty

PR #166, opened 2024-09-19
feat(lib): add fallback to namespace kas

mkleene:
If there is no KAS specified for an attribute value, or attribute, then use the KAS specified on the namespace
to mediate access to the key.

Also refactor and clarify that we need attributes that have been hydrated by the service to provide a split plan.

Addresses https://github.com/virtru-corp/data-security-platform/issues/609


sdk/src/test/java/io/opentdf/platform/sdk/AutoconfigureTest.java
```diff
@@ -426,31 +451,29 @@ public void testReasonerConstructAttributeBoolean() {
...
+                        return String.valueOf(wrapper.i++ + 1);
+                    }
 
-                );
-                assertThat(plan.size()).isEqualTo(tc.getPlan().size());
-                assertThat(plan).isEqualTo(tc.getPlan());
-            });
+            );
+            assertThat(plan.size()).isEqualTo(tc.getPlan().size());
+            assertThat(plan).isEqualTo(tc.getPlan());
```
dmihalcik-virtru, line 476:
```suggestion
            assertThat(plan).isEqualTo(tc.getPlan());
```

dmihalcik-virtru (dismissed):

elizabethhealy (approved):

PR #178, opened 2024-10-04
fix: GitHub packages snapshot repo

pflynn-virtru:
The `<modules>` section was removed from the `pom.xml` file to streamline project configuration; it is defined in profile for each.
The snapshot repository was enabled under the GitHub Packages repository configuration. This ensures better management of snapshot and release artifacts.


elizabethhealy (approved):
does maven central not support snapshots?

PR #184, opened 2024-10-07
chore: release 0.7.2

pflynn-virtru:
Release-As: 0.7.2


mkleene (approved):

mkleene:
It looks like the versions in the pom files will stay at `0.7.2-SNAPSHOT`. Do you think that will work?

PR #187, opened 2024-10-08
feat(sdk): Add and expose tamper error types

elizabethhealy:
Exception type for tamper detection -- make it the parent to the exception types we want to be able to catch for tamper


sdk/src/main/java/io/opentdf/platform/sdk/KASClient.java
```diff
@@ -182,9 +186,21 @@ public byte[] unwrap(Manifest.KeyAccess keyAccess, String policy) {
...
+            var wrappedKey = response.getEntityWrappedKey().toByteArray();
+            return decryptor.decrypt(wrappedKey);
+        } catch (StatusRuntimeException e) {
+            if (e.getStatus().getCode() == Status.Code.INVALID_ARGUMENT) {
+                // 400 Bad Request
+                throw new KasBadRequestException("rewrap request 400: " + e.toString());
+            } else {
+                // Other errors
+                throw e;
+            }
```
dmihalcik-virtru, line 201:
```suggestion
            }
            throw e;
```

dmihalcik-virtru (approved):

pflynn-virtru (approved):

PR #213, opened 2024-11-22
fix(sdk): remove hex encoding

mkleene:
* add a `tdfVersion` attribute to the manifest
* when loading older versions hex-encode hashes
* Adds `--version` option to command line tool


sujankota (approved):

sdk/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>gson</artifactId>
             <version>2.11.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.vdurmont</groupId>
+            <artifactId>semver4j</artifactId>
```
dmihalcik-virtru, line 49: I don't think you need to add a dep to do `TDF_VERSION + "+" + escape(buildMetadata)` (and the reverse)

sdk/src/main/java/io/opentdf/platform/sdk/Config.java
```diff
@@ -229,6 +230,10 @@ public static Consumer<TDFConfig> withMetaData(String metaData) {
         return (TDFConfig config) -> config.metaData = metaData;
     }
 
+    public static Consumer<TDFConfig> withBuildMetadata(String buildMetadata) {
```
dmihalcik-virtru, line 233: How is/will this be called?

dmihalcik-virtru (approved):

PR #214, opened 2024-11-27
fix(sdk): Fuzz testing and protocol fixes

### LINKED ISSUE #168: We should also fix the DOS issues with segment size
I think we just need to duplicate the work done here: https://github.com/opentdf/platform/pull/1536

jentfoo:
This PR includes a variety of fixes found with fuzz testing.  This PR is likely easiest to consume by reviewing commit by commit.  Here is a highlevel of the changes:
* Protocol Exception improvements - `NullPointerExceptions` have been converted to other exception types.  Please provde feedback on if other exception types should be used for these cases.  In `Fuzzing.java` now also serves to define in testing what exception types are acceptable.  The `catch` specifically lists the types of exceptions that were discovered for each API call, and `throws` are checked exceptions that are not expected to be possible.  As a future improvement we may want to refine this list further and better document what exceptions happen under what conditions.  For now I thought it was best to start with just the `NullPointerException` cases.  Since these cases were numerous, these changes span multiple commits, with each commit focused on a specific area of the protocol.
* Protocol DoS Fixes - The only memory consumption issue discovered was the [counterpart found in the go sdk](https://github.com/opentdf/platform/pull/1536).  A matching fix with the same defaults was implemented here in the java sdk.
* Finally the testing itself is added as `Fuzzing.java` executed through `sdk/fuzz.sh`.  This script is long running, and there are occasional Jazzer failures which are not believed to be real deficiencies (timeouts when `.position()` is called on the stream).  For that reason this testing needs to be done manually, and not expected to be included in CI
* A few optimizations and clarity improvements were also included, as they were noticed while generally trying to get familiar with the codebase.

I look forward to questions and recommendations, thank you!

PR closes #168


pflynn-virtru (approved):

sdk/src/main/java/io/opentdf/platform/sdk/ZipReader.java
```diff
@@ -184,8 +193,8 @@ public int read() throws IOException {
                         return -1;
                     }
                     setChannelPosition();
-                    while (buf.position() != buf.capacity()) {
-                        if (zipChannel.read(buf) < 0) {
+                    while (buf.hasRemaining()) {
+                        if (zipChannel.read(buf) <= 0) {
```
mkleene, line 197: This probably isn't a big deal in the cases we are dealing with here but according to the docs for `read` the channel can return 0 bytes before it's been exhausted.

jentfoo: Thank you for calling this out, this is a discreet CPU usage DoS concern.  You're correct that a zero could be returned for non-blocking streams.  Blocking implementations are guaranteed to block until at least one byte is available, or the end of the stream is reached.

The non-blocking behavior where a zero is potentially returned, results in a tight loop here.  Right now this implementation is not well designed for non-blocking implementations.  And since a tight loop is never ideal, I feel like this change to only support blocking streams is the safest and likely least impactful.

Let me know if you have any thoughts on how this can be handled better.  Thank you!

mkleene: Yeah, I think it make sense to trade off correctness for susceptibility to DoS here. In either case it seems like a lot of work to create something that is both seekable and blocks outside of NFS or something so this doesn't seem to likely.

PR #240, opened 2025-04-16
chore(sdk): Pins protocol buffer sources to tags

dmihalcik-virtru:
- Includes dispatchable workflow to generate PRs that bump this, to be used from the source repo (platform)
- Should these be shas instead of tags?


dmihalcik-virtru:
Closed in favor of #243

PR #243, opened 2025-04-17
chore(ci): Adds workflow to bump protocol buffers

dmihalcik-virtru:
Adds a dependabot-style workflow, `update-platform-branch`, which will bump the protocol buffer source tag, used during the generate-sources build phase for RPCs with the policy and key access services.

It runs on a schedule to create a dependabot/rennovate style PR,  but if you need to bump faster it supports manual dispatch, or, if we get it setup, repository dispatch to make it slightly more automated

### DIFF
--- a/.github/workflows/update-platform-branch.yaml
+++ b/.github/workflows/update-platform-branch.yaml
@@ -1,0 +1,65 @@
+name: "Update Platform Branch"
+
+on:
+  workflow_call:
+    inputs:
+      tag:
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "The new tag or branch to update the platform.branch property to ."
+        required: true
+        default: "protocol/go/v0.2.29"
+
+jobs:
+  update-platform-branch:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout java-sdk repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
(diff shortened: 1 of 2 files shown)


.github/workflows/update-platform-branch.yaml
```diff
@@ -0,0 +1,104 @@
...
+            echo "No release notes found for tag $TAG."
+          fi
+          echo "RELEASE_NOTES=$RELEASE_NOTES" >> $GITHUB_ENV
+          echo "RELEASE_URL=https://github.com/opentdf/platform/releases/tag/$TAG" >> $GITHUB_ENV
+
+      - name: Commit changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b update-platform-branch
```
elizabethhealy, line 81: should the created branch be unique in case the PR hasnt been merged? i would assume if there were already changes on update-platform-branch you may run into merge conflict or something when commiting?

elizabethhealy (dismissed):
other than the above question lgtm

.github/workflows/update-platform-branch.yaml
dmihalcik-virtru: It probably doesn't make sense to have more than one of these active at a time. I'm pretty sure the checkout here will fail if it exists? So we should check for an existing PR and update that instead, yeah

dmihalcik-virtru: Better?

elizabethhealy (approved):

PR #258, opened 2025-05-16
fix(sdk): add coverage

mkleene:
run coverage so that when we upload to sonar we get coverage numbers
* don't have a separate coverage step
* set `maven.compiler.release` to 11 to ensure compatibility the 11 JDKs and APIs
* update the JDK we use in tests to 17 so that sonar works correctly. This shouldn't change anything about
the published library since we still use Java 11 to do that

once we modify code in `examples` or our generated code we will probably have to make some changes
with exclusions but for now this is probably ok.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -74,11 +74,6 @@ jobs:
           java-version: "17"
           distribution: "adopt"
           server-id: github
-      - name: Maven Verify
-        run: mvn --batch-mode verify
-        env:
-          BUF_INPUT_HTTPS_USERNAME: opentdf-bot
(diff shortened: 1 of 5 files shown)


dmihalcik-virtru (approved):
Are we okay going to 17 with downstream integrators?

mkleene:
> Are we okay going to 17 with downstream integrators?

I don't think that we really do. We just use a newer SDK and JRE when we run it and only in the tests.

mkleene:
> Are we okay going to 17 with downstream integrators?

I also set the compiler release to ensure we don't use any features/apis that don't exist in 11

dmihalcik-virtru (approved):

PR #364, opened 2026-05-07
feat(sdk)!: remove the hardcoded `BouncyCastle` provider

mkleene:
## Summary by CodeRabbit

* **Refactor**
  * Migrated cryptographic operations to use standard Java security interfaces instead of third-party provider-specific implementations, improving compatibility and reducing external dependencies.
  * Updated EC key handling and validation logic to align with standard cryptographic interfaces.

* **Tests**
  * Enhanced test infrastructure for cryptographic provider initialization and improved test reliability with corrected assertions and variable naming.


marythought (changes requested):
## DX / docs review

Thanks for taking this on — letting consumers bring their own provider is the right call for FIPS. A few things I'd like to see addressed before merge, mostly around making the breaking change discoverable for downstream users.

### 1. PR description needs a migration note

Right now the body is just CodeRabbit's auto-summary. Because Release Please auto-generates the CHANGELOG from the PR title and links back to this PR, the description **is** the migration doc downstream consumers will land on when they hit the `⚠ BREAKING CHANGES` entry in release notes. It needs a hand-written "What breaks / How to migrate" section. Suggested structure:

- **What's changing:** SDK no longer auto-registers `BouncyCastleProvider`.
- **Why:** FIPS compatibility — consumers can now plug in their own provider.
- **Action required:** Register BC (or your preferred provider) in your application's bootstrap, e.g.:
  ```java
  static {
      Security.addProvider(new BouncyCastleProvider());
  }
  ```
  See `cmdline/src/main/java/io/opentdf/platform/Command.java` for the reference pattern.

### 2. There are public API removals beyond provider registration

A few changes go further than the title suggests, and I think they deserve explicit callouts in the migration note:

- **`ECKeyPair.getPEMPublicKeyFromX509Cert(String)` is deleted entirely.** Is this intentional? If anything external depends on it, we should either restore it (using the new BC-classes-without-provider pattern) or document the removal + replacement.
- **`publicKeyFromPem` / `privateKeyFromPem`** now return `java.security.interfaces.ECPublicKey` / `ECPrivateKey` instead of the BC interfaces. Any caller invoking BC-specific methods like `.getQ()` will fail to compile — the deleted `extractPemPubKeyFromX509` test is the canonical example of this pattern.
- **`compressECPublickey(String)` no longer throws `NoSuchProviderException`.** Minor, but a checked-exception signature change.

### 3. README "Cryptography Library" section needs updating

The current wording hedges:

> Note: When using this SDK, it may be necessary to register the Bouncy Castle Provider as follows...

That was accurate when the SDK auto-registered. After this PR it's misleading. I'd update it to:

- State clearly that the SDK no longer auto-registers a provider.
- Spell out when registration is required vs. optional, and what the JDK default supports.
- Mention the FIPS motivation so FIPS users know they can plug in their own provider.
- Point at `Command.java` as a reference pattern.

I'd also verify the Quick Start example in the README still runs end-to-end on a stock JDK after this change. If it doesn't, the example needs the registration block added so first-time users aren't met with a runtime failure.

### 4. opentdf.io docs follow-up

Not in scope for this PR, but I'll audit these for samples that import `org.bouncycastle.jce.interfaces.ECPublicKey` or call `getPEMPublicKeyFromX509Cert`, and update them once this lands:

- https://opentdf.io/sdks/tdf (Java tab)
- https://opentdf.io/sdks/policy (Java tab)
- https://opentdf.io/quickstart (Java sections)
- https://opentdf.io/llms.txt

### What I liked

The test infrastructure (`CryptoProviderSetupExtension` + `META-INF/services` + `junit.jupiter.extensions.autodetection.enabled`) is a clean way to keep BC available for tests without coupling the SDK to it. Nice contributor DX.

mkleene:
> ## DX / docs review
> 
> Thanks for taking this on — letting consumers bring their own provider is the right call for FIPS. A few things I'd like to see addressed before merge, mostly around making the breaking change discoverable for downstream users.
> ### 1. PR description needs a migration note
> 
> Right now the body is just CodeRabbit's auto-summary. Because Release Please auto-generates the CHANGELOG from the PR title and links back to this PR, the description **is** the migration doc downstream consumers will land on when they hit the `⚠ BREAKING CHANGES` entry in release notes. It needs a hand-written "What breaks / How to migrate" section. Suggested structure:
> 
>     * **What's changing:** SDK no longer auto-registers `BouncyCastleProvider`.
> 
>     * **Why:** FIPS compatibility — consumers can now plug in their own provider.
> 
>     * **Action required:** Register BC (or your preferred provider) in your application's bootstrap, e.g.:
>       ```java
>       static {
>           Security.addProvider(new BouncyCastleProvider());
>       }
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       See `cmdline/src/main/java/io/opentdf/platform/Command.java` for the reference pattern.
> 
> 
> ### 2. There are public API removals beyond provider registration
> 
> A few changes go further than the title suggests, and I think they deserve explicit callouts in the migration note:
> 
>     * **`ECKeyPair.getPEMPublicKeyFromX509Cert(String)` is deleted entirely.** Is this intentional? If anything external depends on it, we should either restore it (using the new BC-classes-without-provider pattern) or document the removal + replacement.
> 
>     * **`publicKeyFromPem` / `privateKeyFromPem`** now return `java.security.interfaces.ECPublicKey` / `ECPrivateKey` instead of the BC interfaces. Any caller invoking BC-specific methods like `.getQ()` will fail to compile — the deleted `extractPemPubKeyFromX509` test is the canonical example of this pattern.
> 
>     * **`compressECPublickey(String)` no longer throws `NoSuchProviderException`.** Minor, but a checked-exception signature change.
> 
> 
> ### 3. README "Cryptography Library" section needs updating
> 
> The current wording hedges:
> 
> > Note: When using this SDK, it may be necessary to register the Bouncy Castle Provider as follows...
> 
> That was accurate when the SDK auto-registered. After this PR it's misleading. I'd update it to:
> 
>     * State clearly that the SDK no longer auto-registers a provider.
> 
>     * Spell out when registration is required vs. optional, and what the JDK default supports.
> 
>     * Mention the FIPS motivation so FIPS users know they can plug in their own provider.
> 
>     * Point at `Command.java` as a reference pattern.
> 
> 
> I'd also verify the Quick Start example in the README still runs end-to-end on a stock JDK after this change. If it doesn't, the example needs the registration block added so first-time users aren't met with a runtime failure.
> ### 4. opentdf.io docs follow-up
> 
> Not in scope for this PR, but I'll audit these for samples that import `org.bouncycastle.jce.interfaces.ECPublicKey` or call `getPEMPublicKeyFromX509Cert`, and update them once this lands:
> 
>     * https://opentdf.io/sdks/tdf (Java tab)
> 
>     * https://opentdf.io/sdks/policy (Java tab)
> 
>     * https://opentdf.io/quickstart (Java sections)
> 
>     * https://opentdf.io/llms.txt
> 
> 
> ### What I liked
> 
> The test infrastructure (`CryptoProviderSetupExtension` + `META-INF/services` + `junit.jupiter.extensions.autodetection.enabled`) is a clean way to keep BC available for tests without coupling the SDK to it. Nice contributor DX.

I'm not sure what the right pattern is for the provider. It seems like telling them they can bring their own provider is wrong because we use bouncycastle internally, e.g. [this](https://github.com/opentdf/java-sdk/blob/9991b077f7cd5a3c5fc1b0cd1a5d1ca1abe10b96/sdk/src/main/java/io/opentdf/platform/sdk/ECKeyPair.java#L60-L64). We need a static binding to bouncycastle but at that point we should just load the provider for them, I think, since if we are statically bound to one provider and they pass us the other then we are in trouble. Maybe it's better to just depend on the bouncycastle classes explicitly and not go through the provider interfaces? I wonder if the right thing to do is just list bouncycastle as `provided` and then make sure that a BC provider is registered. Not great since forgetting to do that would result in a ClassNotFoundException, which is pretty ugly.

Ideally there would be a way to parse the PEM stuff in a way that is unrelated to FIPS. We can't use https://openjdk.org/jeps/524 until java 26, which kind of seems too much to me.

I also think I made a mistake in opening this without removing the `ayza` dependencies, since they explicitly load a non-FIPS bouncycastle provider, which kind of precludes loading the FIPS provider.

PR #367, opened 2026-05-18
feat(sdk)!: remove bouncycastle and ayza libraries

mkleene:
This pull makes this library provider-agnostic. Users can control how cryptography is provided by configuring
their `java.security` configuration, as we do in the FIPS-profile tests. We add two new maven profiles `fips` and
`non-fips`. These profiles are used solely for testing; the library itself is agnostic to which providers are used
for cryptography.

In the `non-fips` profile we use the normal providers configured on the JVM. The BouncyCastle provider is included
just for some test verification.

In the `fips` profile we include the BouncyCastle FIPS providers we need at runtime to verify that our library works with a  FIPS-compliant setup.

## Ayza removal
I believe it's necessary to remove ayza since it loads a BouncyCastleProvider explicitly. Since the FIPS provider
occupies the same namespace loading both providers can't work.

## BouncyCastle moved to test-only
Removing BouncyCastle as a compile dependency isn't strictly necessary but it results in much less rework and
eliminates a transitive dependency for consumers.

## Remaining FIPS work
We still need to:
* verify that the JWK libraries we  use are FIPS compliant with the correct providers to claim that this
library can be made FIPS compliant
* verify that we generate random numbers in a FIPS-compliant way. Because BouncyCastle does not provide a
FIPS-compliant source of entropy we need to include the SUN provider to get at `/dev/rand`. This is OK, since
  * we _should_ pick up this provider and use it for all cryptographic operations (but we may want to provide ways to guarantee that a particular provider is used)
  * the FIPS boundary for random numbers starts at the provider. We don't have tools to verify this now, though

The breaking API changes are as follows:

| Area | Change | Related to |
|---|---|---|
| `SDKBuilder.sslFactory(SSLFactory)` | **Removed.** Replaced by `sslFactory(SSLSocketFactory)` and `sslFactory(SSLSocketFactory, X509TrustManager)` | Ayza removal |
| `SDKBuilder.sslFactoryFromTrustManager(X509TrustManager)` | **New** method | Ayza removal |
| `ECKeyPair(ECCurve, ECAlgorithm)` constructor | **Removed.** Replaced by `ECKeyPair(ECCurve)` | BouncyCastle removal |
| `ECKeyPair.publicKeyFromPem` | Removed method from public interface since it should not have been exposed and since the types were changing anyway | BouncyCastle removal |
| `ECKeyPair.privateKeyFromPem`, `getPEMPublicKeyFromX509Cert`, `publicKeyFromECPoint`, static `compressECPublickey(String)` | **Removed** from public API (moved to test-only `PemTestUtils`) | BouncyCastle removal |
| Dependencies | Drops `io.github.hakky54:ayza*` and the runtime BC dep; adds `bc-fips`, `bcpkix-fips`, `bctls-fips` (via new `fips`/`non-fips` Maven profiles) | Both |

## Summary by CodeRabbit

* **New Features**
  * FIPS-profile cryptography support and improved TLS trust-material builder (directory/keystore/trust-manager sources).

* **Improvements**
  * Migrated crypto to standard Java security APIs for broader compatibility.
  * Added AES-GCM key-generation helper and more robust TLS/trust-manager integration across the SDK.

* **Tests**
  * Updated and added tests and test utilities for PEM/EC key handling and TLS/trust behaviors.

* **Chores**
  * Updated build/test config and pinned Bouncy Castle FIPS dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/java-sdk/pull/367?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sdk/src/main/java/io/opentdf/platform/sdk/TrustProvider.java
```diff
@@ -0,0 +1,238 @@
...
+                }
+                builder.withTrustMaterial(x509s.toArray(new X509Certificate[0]));
+            }
+        }
+        return builder.build();
+    }
+
+    public static TrustProvider fromTrustManager(X509TrustManager trustManager) throws GeneralSecurityException {
+        SSLContext sslContext = SSLContext.getInstance("TLS");
+        sslContext.init(new KeyManager[0], new TrustManager[]{trustManager}, new SecureRandom());
```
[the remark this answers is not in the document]
mkleene: in this case we do want an insecure provider

sdk/pom.xml
```diff
@@ -483,11 +467,56 @@
...
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bctls-fips</artifactId>
+                    <scope>runtime</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bcpkix-fips</artifactId>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
```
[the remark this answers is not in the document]
mkleene: these aren't dependencies in the normal sense and they are often installed on the JVM themselves

cmdline/src/main/java/io/opentdf/platform/Command.java
```diff
@@ -262,10 +263,13 @@ void encrypt(
...
-                    .withUnsafeTrustMaterial() // Trust all certificates
-                    .build();
-            builder.sslFactory(sslFactory);
+            // Trust all certificates
+            X509TrustManager insecureTrustManager = new X509TrustManager() {
+                @Override public void checkClientTrusted(X509Certificate[] chain, String authType) {}
+                @Override public void checkServerTrusted(X509Certificate[] chain, String authType) {}
+                @Override public X509Certificate[] getAcceptedIssuers() { return new X509Certificate[0]; }
+            };
+            builder.sslFactoryFromTrustManager(insecureTrustManager);
```
marythought: The insecure-trust path is now ~5 lines of boilerplate every consumer who wants local-dev "trust everything" will rewrite. The pre-ayza one-liner was `SSLFactory.builder().withUnsafeTrustMaterial().build()`. Worth a no-arg `SDKBuilder.useInsecureSkipVerify()` (or similar) that wraps this same anonymous trust-manager internally? `useInsecurePlaintextConnection` is already alongside it as precedent for "we provide the unsafe option explicitly because consumers will ask for it anyway."

sdk/src/main/java/io/opentdf/platform/sdk/ECKeyPair.java
```diff
@@ -1,39 +1,31 @@
...
+import java.security.spec.ECGenParameterSpec;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPoint;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Base64;
 import java.util.Objects;
-// https://www.bouncycastle.org/latest_releases.html
 
 public class ECKeyPair {
```
marythought, line 25: The PR description's breaking-change table lists what's removed from `ECKeyPair`'s public API, but doesn't say what consumers should use instead. For each removed static helper, a one-line JCA recipe in the table would save callers a JCA-docs hunt — e.g., `publicKeyFromPem(pem)` → `KeyFactory.getInstance("EC").generatePublic(new X509EncodedKeySpec(...))`, or `getPEMPublicKeyFromX509Cert(...)` → `CertificateFactory.getInstance("X.509").generateCertificate(...).getPublicKey()`. If any of the removed helpers have no public replacement (genuinely test-only), saying so explicitly avoids a search that ends in disappointment.

sdk/src/main/java/io/opentdf/platform/sdk/TrustProvider.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Builds {@link SSLSocketFactory} and {@link X509ExtendedTrustManager} instances for verifying
+ * X.509 certificate chains during TLS handshakes.
+ *
+ * <p>Implemented entirely on top of provider-agnostic JCA APIs ({@link CertificateFactory},
+ * {@link KeyStore}, {@link TrustManagerFactory}, {@link SSLContext}). The actual cryptographic
+ * work is fulfilled by whichever {@link java.security.Provider} is registered with the JVM,
+ * including FIPS-mode providers.
+ */
+final class TrustProvider {
```
marythought, line 40: Load-bearing architectural shift here — provider-agnostic JCA + an internal trust composition abstraction, driven by FIPS-readiness and removing the BC namespace conflict with ayza. The rationale is solid but lives only in the PR description today. Worth recording in a repo ADR (e.g., `adr/000X-provider-agnostic-cryptography.md`) so future maintainers don't have to dig through PR history to understand why BC stopped being a compile dep and why `TrustProvider` exists. Could also capture the trade-offs that aren't visible in code — e.g., why `TrustProvider` is package-private today, and what the FIPS-readiness gaps are (the "Remaining FIPS work" list in the PR body).

marythought (commented):
Overall — this is a real DX improvement, even with the breaking changes. From the consumer's seat:

- **Dropping ayza is a long-term win.** Removes a niche third-party type from the public surface; `SSLSocketFactory` / `X509TrustManager` are types every Java dev already knows.
- **The new `SDKBuilder` method set is well-shaped.** `sslFactoryFromTrustManager` / `sslFactoryFromDirectory` / `sslFactoryFromKeyStore` cover the common cases; the bare/paired `sslFactory(SSLSocketFactory[, X509TrustManager])` overloads handle the escape hatches; naming is consistent and the javadoc explicitly tells callers to prefer the paired form. Good guidance built into the API.
- **Sensible defaults — `fromDirectory` / `fromKeyStore` include JVM default cacerts automatically.** The common "my certs *plus* system trust" case works without customers having to think about composition.
- **The breaking-change table in the PR description is exactly the format consumers need.** Old → new in one place.
- **ECKeyPair public-surface tightening is principled.** The moved-to-test-only helpers were marked as "should not have been exposed" and would have broken under FIPS anyway. Better to fix now.

Three specific notes as inline comments below.

cmdline/src/main/java/io/opentdf/platform/Command.java
mkleene: That makes sense. I'll add a method like that.

sdk/src/main/java/io/opentdf/platform/sdk/ECKeyPair.java
mkleene: I think that if they are looking for a recipe the pointer to the test class should be OK, no? There's no real reason for anyone to use this in the context of TDF so I think that should be all they'd need.

sdk/src/main/java/io/opentdf/platform/sdk/TrustProvider.java
mkleene: I've added an ADR (that doesn't address some of the things you mentioned) but I'd think that the burden of why should almost always come out in the other direction; "why is this thing public?"

marythought: agree! in this case it's already public but I like that approach in the future

marythought (approved):
Re-reviewed after the latest commits. Both items from my first review are addressed:

1. `insecureSslFactory()` convenience method added to `SDKBuilder` — clean one-liner for local-dev consumers.
2. ADR documenting the BouncyCastle/Ayza removal rationale.

CI is green on all test/integration/xtest jobs. The SonarCloud and dependency-review failures look infrastructure-related.

LGTM — nice work on the provider-agnostic JCA migration.

jentfoo (approved):
LGTM

PR #368, opened 2026-05-18
feat(sdk): DSPX-3309 add hybrid post-quantum key wrapping for KAS (X-Wing, ECDH+ML-KEM)

sujankota:
Add hybrid post-quantum key wrapping to the Java SDK so TDFs can be protected against "harvest now, decrypt later" attacks while preserving classical security guarantees during the PQC transition.
  
  Introduces three new `KeyType` values backed by hybrid KEMs:

  - `HybridXWingKey` (`hpqt:xwing`) — X-Wing (X25519 + ML-KEM-768)
  - `HybridSecp256r1MLKEM768Key` (`hpqt:secp256r1-mlkem768`)
  - `HybridSecp384r1MLKEM1024Key` (`hpqt:secp384r1-mlkem1024`)

  When a KAS advertises one of these algorithms,`TDF.upsertAndGetNewKeyAccess` routes through `HybridCrypto.wrapDEK`, which performs both a classical ECDH/X25519 key agreement and an ML-KEM encapsulation, combines the two shared secrets (HKDF-SHA256 with the standard TDF salt), and uses the result to wrap the DEK with AES-256-GCM. A new `hybrid-wrapped` key-access type is emitted; the ephemeral classical public key and ML-KEM ciphertext are packaged together inside an ASN.1 envelope stored in `wrappedKey` (rather than the separate `ephemeralPublicKey` field used for EC-wrapped keys).

  New supporting classes: `HybridCrypto`, `HybridNISTKeyPair`, `XWingKeyPair`, plus unit tests and a full-manifest TDF round-trip test.

  ## Provider-agnostic implementation 

  In line with #367's removal of BouncyCastle as a compile dependency, this PR limits BC usage to the only primitives JDK 11 stdlib cannot supply — ML-KEM keygen/encap/decap and X-Wing keygen/encap/decap (the JCA `KEM` API is JDK21+; ML-KEM in stdlib is 24+).

  Everything else is stdlib JCA or an existing SDK helper:

  - **ASN.1 envelope** — minimal hand-rolled DER codec for `SEQUENCE { [0] IMPLICIT OCTET STRING, [1] IMPLICIT OCTET STRING }` with multi-byte length support. No `org.bouncycastle.asn1.*` imports. - **HKDF** — delegates to the existing `ECKeyPair.calculateHKDF(salt, secret)` (RFC 5869, empty info, L=32 — what all three algorithms need).
  - **EC keygen / ECDH / curve parameters** — stdlib
    `KeyPairGenerator.getInstance("EC")`,
    `KeyAgreement.getInstance("ECDH")`,
    `AlgorithmParameters.getInstance("EC")` with `ECGenParameterSpec`. No
    `BouncyCastleProvider` registration; consumers control providers via
    `java.security` as the ADR intends.

  `bcprov-jdk18on` is declared at compile/runtime scope in the default
  `non-fips` Maven profile, version pinned via the existing parent
  dep-management entry.

  ## Out of scope (follow-ups)

  - **`fips` profile support for hybrid PQC** — needs verification of which
    `bc-fips` version ships ML-KEM and X-Wing and how it registers them.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/java-sdk/pull/368?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Release Notes

* **New Features**
  * Added support for hybrid post-quantum cryptography with three new key types: X-Wing, secp256r1+ML-KEM-768, and secp384r1+ML-KEM-1024.
  * Introduced hybrid key wrapping mechanism for TDF encryption, enabling classical and post-quantum algorithm combinations.

* **Tests**
  * Added end-to-end testing suite for hybrid post-quantum key wrapping workflows.

* **Documentation**
  * Updated test documentation with hybrid PQC integration workflow and troubleshooting guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/java-sdk/pull/368?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jentfoo (dismissed):
No issues spotted, just unfortunate having to bring BC back after [just removing it](https://github.com/opentdf/java-sdk/pull/367).

mkleene (commented):
I think that this breaks FIPS, at least with BouncyCastle. I'm not sure which other providers are out there but it doesn't seem like there are many. 

Since these classes aren't present in the FIPS jar we'd need to include the regular BC jar in order to guarantee that we don't have link errors. The [spec](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-5.html#jvms-5.4) says

> 
> This specification allows an implementation flexibility as to when linking activities (and, because of recursion, loading) take place, provided that all of the following properties are maintained:
> 
>     A class or interface is completely loaded before it is linked.
> 
>     A class or interface is completely verified and prepared before it is initialized.
> 
>     Errors detected during linkage are thrown at a point in the program where some action is taken by the program that might, directly or indirectly, require linkage to the class or interface involved in the error.
> 
>     A symbolic reference to a dynamically-computed constant is not resolved until either (i) an ldc, ldc_w, or ldc2_w instruction that refers to it is executed, or (ii) a bootstrap method that refers to it as a static argument is invoked.
> 
>     A symbolic reference to a dynamically-computed call site is not resolved until a bootstrap method that refers to it as a static argument is invoked.

which makes me think that a class loader would be justified in throwing a ClassNotFoundException at startup if the regular BC provider is not on the class path. But if the regular provider is on the classpath then the FIPS jar will not work properly because of collisions in the standard and FIPS class names.

If we really want this to work we could put it behind a separate jar and the java SPI maybe?

scripts/test-hybrid-pqc.sh
mkleene, line 1: can we turn this into a test that runs in CI like [this](https://github.com/opentdf/java-sdk/blob/63b49af858f5eb78ea3db72d2fe991abdb3fcec8/.github/workflows/checks.yaml#L140-L173)?

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -226,7 +228,16 @@ private Manifest.KeyAccess createKeyAccess(Config.TDFConfig tdfConfig, Config.KA
                     : kasInfo.Algorithm;
 
             var keyType = KeyType.fromString(algorithm);
-            if (keyType.isEc()) {
+            if (keyType.isHybrid()) {
+                // Dispatch to whichever KemProvider claims this KeyType (typically the
+                // BouncyCastle-backed impl in sdk-pqc-bc). Keeps the core sdk jar free
+                // of BC compile-time references so the fips Maven profile stays clean.
+                byte[] wrapped = KemProviders.get(keyType).wrapDEK(keyType, kasInfo.PublicKey, symKey);
```
mkleene, line 235: maybe when this fails give them some kind of better error message with something the user can do to fix this?

mkleene (commented):
Overall I think this will work well. Keeps the FIPS story pretty simple and should be simple to consume for customers.

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
mkleene: Found what happens when one is missing and it looks good

marythought (commented):
## DX Review

Strong PR — the SPI + ServiceLoader architecture is clean, the test script is genuinely helpful, and the wire format is well-documented for cross-SDK interop. A few things worth considering:

### Design feedback

1. **`HybridNISTKeyPair` dual role.** `P256_MLKEM768` is a static template with `publicKey = null`, and `.generate()` returns a new instance with keys populated — but both are the same type. Nothing prevents calling instance methods on the template, and the API shape suggests `P256_MLKEM768` itself holds a keypair. Consider separating the parameters object from the keypair, or making the distinction visible in the type name.

2. **Duplicated dispatch switch.** `HybridCrypto.wrapDEK` and `BouncyCastleKemProvider.wrapDEK` both switch over the same three `KeyType`s dispatching to the same calls. A fourth algorithm means two switch statements to update. Consider having the provider delegate to `HybridCrypto`, or removing the dispatcher from `HybridCrypto` entirely.

3. **Auto-discovery gap deserves a defensive error now.** The README documents that `fromPublicKeyAlgorithm()` doesn't map hybrid protobuf enums. When unrecognized enum values do arrive, the current `default` case throws a bare `IllegalArgumentException` with no mention of PQC or what to do. Since the enum mapping can't be added until the protobuf values exist, the fix available now is making `fromPublicKeyAlgorithm()` handle unknown values gracefully — either skip with a log warning, or throw an error that says what's happening. Either is better than a raw `IllegalArgumentException`.

### Minor

- `concat()` is defined in three places (`HybridCrypto`, `HybridNISTKeyPair`, `XWingKeyPair`).
- `HybridCrypto.defaultTDFSalt()` recomputes `SHA-256("TDF")` on every call — `TDF.java` already has `GLOBAL_KEY_SALT` doing the same thing.
- `XWingKeyPair.PRIVATE_KEY_SIZE = 32` — worth a note that this is a seed, not the full private key.
- `sdk-pqc-bc/pom.xml` hardcodes `<maven.compiler.source>11</maven.compiler.source>` — intentional divergence from parent?
- `KemProvider.supportedKeyTypes()` has no contract about immutability/thread-safety of the returned `Set` — matters since this is a public SPI.

pom.xml
mkleene, line 1: I think it's ok for this to work this way but it might be simpler to do dependency management inside the commandline module in order to keep things out. That said, I think that this is fine and will work.

mkleene (dismissed):
I think this is fine to merge. It would be nice if we had a platform integration test but it looks like the tests repo should handle e2e tests.

Looks good!

marythought (commented):
**Re: Mike's concern about bringing BC back after #367 just removed it** — the SPI + ServiceLoader architecture addresses this correctly. The core `sdk` jar stays BC-free at compile time; `sdk-pqc-bc` is an optional module gated behind the `non-fips` profile. The FIPS namespace collision Mike flagged is avoided because `sdk-pqc-bc` is excluded from the reactor entirely under `-P fips,!non-fips`.

One remaining gap: FIPS users who attempt hybrid PQC get a generic "no KemProvider registered — add sdk-pqc-bc" error, which is misleading because adding `sdk-pqc-bc` under FIPS would *create* the namespace collision Mike described. A more targeted message when running under the FIPS profile (or at least a note that hybrid PQC is not available in FIPS mode) would prevent someone from following the error message into a worse state.

sdk-pqc-bc/src/main/java/io/opentdf/platform/sdk/pqc/bc/HybridNISTAlgorithm.java
```diff
@@ -0,0 +1,447 @@
...
+            ECPoint w = scalarMultiplyGenerator(d);
+            return encodeUncompressedPoint(w, ecFieldByteSize);
+        } catch (Exception e) {
+            throw new SDKException("failed to recover EC public point on " + curveName, e);
+        }
+    }
+
+    /** Plain double-and-add scalar multiplication of the curve generator. */
+    private ECPoint scalarMultiplyGenerator(BigInteger k) {
+        java.security.spec.EllipticCurve curve = ecParams.getCurve();
```
marythought: This is a non-constant-time double-and-add implementation operating on a private scalar. For an unwrap path (private key operation), timing side channels on the scalar are a real concern.

BouncyCastle is already on the classpath in this module — `org.bouncycastle.math.ec.ECPoint.multiply()` is constant-time and would replace these ~50 lines with ~3. If there's a reason to avoid that BC API here, a comment acknowledging the timing risk as accepted would be good.

sdk/src/main/java/io/opentdf/platform/sdk/KeyType.java
```diff
@@ -14,7 +14,10 @@ public enum KeyType {
     RSA4096Key("rsa:4096"),
     EC256Key("ec:secp256r1", SECP256R1),
     EC384Key("ec:secp384r1", SECP384R1),
-    EC521Key("ec:secp521r1", SECP521R1);
+    EC521Key("ec:secp521r1", SECP521R1),
+    HybridXWingKey("hpqt:xwing"),
+    HybridSecp256r1MLKEM768Key("hpqt:secp256r1-mlkem768"),
```
marythought, line 29: These hybrid key types require `sdk-pqc-bc` on the classpath at runtime — without it, the user gets a runtime `SDKException` from `KemProviders.get()` with no compile-time signal. A short Javadoc note on at least one of these values (e.g., "Requires the `sdk-pqc-bc` module or another `KemProvider` implementation") would save someone a debugging round-trip.

marythought:
Companion docs PR: https://github.com/opentdf/docs/pull/343 — updates feature matrix, encrypt options, KASInfo/KeyAccess type reference, and KAS PublicKey docs for hybrid PQC.

sdk-pqc-bc/src/main/java/io/opentdf/platform/sdk/pqc/bc/HybridNISTAlgorithm.java
marythought: Tracked in [DSPX-3626](https://virtru.atlassian.net/browse/DSPX-3626) — not blocking merge since this path is test-only today (production decrypt defers to KAS).

sdk/src/main/java/io/opentdf/platform/sdk/KeyType.java
marythought: Tracked in [DSPX-3626](https://virtru.atlassian.net/browse/DSPX-3626).

marythought (dismissed):

sujankota:
@copilot resolve the merge conflicts in this pull request

dmihalcik-virtru (dismissed):

mkleene (approved):

PR #375, opened 2026-06-09
feat: verify that BC works with `approved_only` mode set

mkleene:
The bouncycastle FIPS provider happily does things that are not FIPS-compliant when the
`org.bouncycastle.fips.approved_only` system property is not set to `true`. The following problems were observed:

1. using OAEP to `ENCRYPT` rather than `WRAP`. We are indeed using OEAP for key wrapping so we just had
to change some cipher parameters. `WRAP` and `ENCRYPT` perform the same operation; the only difference
is in the semantics and the JCA API associated with each operation. This should not affect compatibility with other parts of the platform in any way.
1. Without `approved_only` BC would be fine using a non-FIPS truststore. Now we add an empty truststore
(with an empty password) of a type that can be loaded by the BC FIPS truststore.
1. Our home grown implementation of HKDF didn't work because our usage of `SHA-265/HMAC` within the HKDF
implementation triggers the error: `ECKeyPairTest.createSymmetricKeysWithOtherCurves:129 » IllegalKey Key size for HMAC must be at least 112 bits in approved mode: SHA-256/HMAC`. By using a FIPS compliant provider that implements HKDF directly we bypass this error.

We include the FIPS implementation of HKDF in a separate JAR due to its dependency on the bouncycastle FIPS provider. If we included it in the same JAR we would have a dependency on `bc-fips`, which would conflict with usage of the regular bouncycastle provider. Since bouncycastle is popular it seems best to allow non-FIPS usage.

The major change from a customer POV is that customers that want to use EC crypto in a FIPS context will now
need to pull in the `sdk-fips-bouncycastle` jar and depend on `bc-fips` since HKDF [only got a JCA name
in JDK 24](https://openjdk.org/jeps/478) and [bc-fips does not expose HKDF through a provider interface](https://downloads.bouncycastle.org/fips-java/docs/BC-FJA-UserGuide-2.1.1.pdf). If using `bc-fips` is not appropriate
in a particular application it would be quite easy to implement an alternative JAR that implements a similar
provider pattern and uses whichever FIPS-compliant library one wishes to implement HKDF. 

Also:
* pull the test scripts in `checks.yaml` in to bash files so they can be run locally more easily
* add a FIPS mode `cmdline` platform integration phase
* publish the new FIPS provider JAR to maven central
* remove double addition of generated sources that was generating a warning
* overwrite the `java.security` file in its entirety so that we know exactly which providers we are using in FIPS tests


pflynn-virtru (dismissed):
Nice to have an interop test with FIPS and non-FIPS

marythought (commented):
## DX Review — release / migration impact

The FIPS fixes are correct, but this is a breaking change for existing FIPS + EC deployments and the release process needs to surface that.

### 1. Migration note via release-please

Since this repo uses release-please, the squash-merge commit message needs a `BREAKING CHANGE:` footer so it appears in the generated release notes. Something like:

```
BREAKING CHANGE: FIPS deployments using EC key wrapping must add
`io.opentdf.platform:sdk-fips-bouncycastle` to their classpath.
Without it, HKDF key derivation fails under
`org.bouncycastle.fips.approved_only=true`.
```

Without this footer, the change ships as a minor `fix:` and FIPS consumers discover the new dependency requirement at runtime.

### 2. Failure mode isn't discoverable

If a FIPS consumer doesn't add the jar: `HkdfResolver.get()` returns null → `calculateHKDF` falls through to the JDK-native HmacSHA256 → same `IllegalKey` error as before. The error doesn't mention `sdk-fips-bouncycastle` or how to fix it. Consider logging a warning when `approved_only=true` is set but no `HkdfProvider` resolves — that would point people to the fix instead of leaving them with a raw BC exception.

### 3. Wire compatibility of ENCRYPT→WRAP cipher mode change

The `AsymEncryption`/`AsymDecryption` change from `ENCRYPT_MODE`/`DECRYPT_MODE` to `WRAP_MODE`/`UNWRAP_MODE` should be wire-compatible (OAEP produces the same bytes either way), but worth confirming in the release notes that existing TDFs encrypted with the old mode can still be decrypted — this is the kind of thing a security-conscious consumer will ask about.

sujankota (dismissed):

mkleene:
> ## DX Review — release / migration impact
> 
> The FIPS fixes are correct, but this is a breaking change for existing FIPS + EC deployments and the release process needs to surface that.
> ### 1. Migration note via release-please
> 
> Since this repo uses release-please, the squash-merge commit message needs a `BREAKING CHANGE:` footer so it appears in the generated release notes. Something like:
> 
> ```
> BREAKING CHANGE: FIPS deployments using EC key wrapping must add
> `io.opentdf.platform:sdk-fips-bouncycastle` to their classpath.
> Without it, HKDF key derivation fails under
> `org.bouncycastle.fips.approved_only=true`.
> ```
> 
> Without this footer, the change ships as a minor `fix:` and FIPS consumers discover the new dependency requirement at runtime.
> ### 2. Failure mode isn't discoverable
> 
> If a FIPS consumer doesn't add the jar: `HkdfResolver.get()` returns null → `calculateHKDF` falls through to the JDK-native HmacSHA256 → same `IllegalKey` error as before. The error doesn't mention `sdk-fips-bouncycastle` or how to fix it. Consider logging a warning when `approved_only=true` is set but no `HkdfProvider` resolves — that would point people to the fix instead of leaving them with a raw BC exception.
> ### 3. Wire compatibility of ENCRYPT→WRAP cipher mode change
> 
> The `AsymEncryption`/`AsymDecryption` change from `ENCRYPT_MODE`/`DECRYPT_MODE` to `WRAP_MODE`/`UNWRAP_MODE` should be wire-compatible (OAEP produces the same bytes either way), but worth confirming in the release notes that existing TDFs encrypted with the old mode can still be decrypted — this is the kind of thing a security-conscious consumer will ask about.

1. I don't think that this is necessary since it never could have worked in approved_only mode. It will continue to work in non-approved mode and now there is a way to get it working in approved_only mode.
2. added a test. well, a disabled test. I can't think of a way to add test that runs without a whole new profile and that seems a bit much for this.
3. This should be covered by our existing `golden-file` tests, at least for decryption. The fact that this works with the unmodified platform verifieds that encryption will also work with older versions of the platform.

sujankota (approved):

PR #388, opened 2026-07-31
feat(sdk): support ML-KEM rewrap session keys

dmihalcik-virtru:
## Summary

Part of DSPX-4221 (Session Keys should support ML-KEM). `KASClient.unwrap()` only branched EC vs RSA for the client's ephemeral rewrap "session key" (as opposed to a KAS-managed wrapping key), so requesting an ML-KEM session key would fail. `KeyType` already had `MLKEM768Key`/`MLKEM1024Key` entries and `sdk-pqc-bc` already had the BouncyCastle KEM primitives (used for the KAS-managed mechanism-key path), but nothing generated a fresh ML-KEM keypair for the session-key role.

## Changes

- **`spi/KemProvider`**: add `generateKeyPair(KeyType)`, returning a new `KeyPairPem` (PEM-encoded SPKI/PKCS#8 pair). The core `sdk` module has no compile-time dependency on BouncyCastle (per ADR 0001), so the actual keypair generation lives in the optional `sdk-pqc-bc` module and is reached via the existing `ServiceLoader`-based `KemProviders` registry — same pattern already used for `wrapDEK`/`unwrapDEK`.
- **`sdk-pqc-bc/HybridCrypto`**: add a `generateKeyPair` dispatch mirroring the existing `wrapDEK`/`unwrapDEK` switch, covering all five supported PQC key types (X-Wing, both NIST hybrids, ML-KEM-768/1024) so the SPI contract stays complete for every type in `supportedKeyTypes()`.
- **`sdk-pqc-bc/BouncyCastleKemProvider`**: implement the new SPI method.
- **`KASClient.unwrap()`**: branch on `sessionKeyType.isMLKEM()`, generating a keypair via `KemProviders.get(...).generateKeyPair(...)` and decrypting the response via `KemProviders.get(...).unwrapDEK(...)`.
- New unit test `HybridCryptoGenerateKeyPairTest` round-trips `generateKeyPair -> wrapDEK -> unwrapDEK` for all five PQC key types.

`cmdline`'s `--rewrap-key-type` already accepts `mlkem:768`/`mlkem:1024` with **no code changes needed**: picocli's enum candidate list is derived from `KeyType#toString()`, which already returns `"mlkem:768"`/`"mlkem:1024"`. Verified by building `cmdline.jar` and checking `help decrypt` output.

## Test plan

```
mvn -pl sdk,sdk-pqc-bc -am test
```
All existing tests plus the new `HybridCryptoGenerateKeyPairTest` (5/5) pass. Also built `cmdline.jar` and confirmed `--rewrap-key-type=mlkem:768` parses correctly (fails downstream only on the expected network/gRPC call, not on argument validation).

## Related work

Companion PRs:
- opentdf/tests#571 — xtest coverage (`test_session_key_mlkem_roundtrip`) gated behind a new `session-key-mlkem` feature flag.
- opentdf/platform#3814 — accepts ML-KEM client session keys in `rewrap`, and adds the equivalent Go SDK branch.

Ref: DSPX-4221

## Summary by CodeRabbit

* **New Features**
  * Added support for ML-KEM session-key unwrapping.
  * Added key-pair generation for supported hybrid and ML-KEM algorithms.
  * Made generated public and private keys available in PEM format.
  * Added ephemeral key-pair support for cryptographic workflows.
  * Updated the `supports` command to report ML-KEM session-key support.

* **Bug Fixes**
  * Improved ML-KEM key validation and end-to-end wrap/unwrap handling.
  * Preserved existing EC and RSA key-handling behavior.

### DIFF
--- a/cmdline/src/main/java/io/opentdf/platform/Command.java
+++ b/cmdline/src/main/java/io/opentdf/platform/Command.java
@@ -43,6 +43,7 @@ import io.opentdf.platform.sdk.Config;
 import io.opentdf.platform.sdk.KeyType;
 import io.opentdf.platform.sdk.SDK;
 import io.opentdf.platform.sdk.SDKBuilder;
+import io.opentdf.platform.sdk.spi.KemProviders;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.core.config.Configurator;
 import picocli.CommandLine;
@@ -75,9 +76,21 @@ class Command {
 
     @CommandLine.Command(name = "supports", description = "Check if a feature is supported, or list all supported features")
     static class Supports implements Callable<Integer> {
-        // Static, always-on capabilities of this build. There is no "known but
+        // Always-on capabilities of this build. There is no "known but
         // unsupported" feature here: anything not in this list is simply unrecognized.
-        private static final List<String> FEATURES = List.of("dpop", "dpop_nonce_challenge");
+        private static final List<String> STATIC_FEATURES = List.of("dpop", "dpop_nonce_challenge");
+
+        // session-key-mlkem additionally requires an ML-KEM KemProvider on the
+        // classpath (e.g. sdk-pqc-bc), which the fips Maven profile excludes.
+        // Check the live registry rather than assuming build-profile wiring, so
+        // this doesn't claim support on a classpath that would actually throw.
+        private static List<String> features() {
+            List<String> features = new ArrayList<>(STATIC_FEATURES);
+            if (KemProviders.registered().contains(KeyType.MLKEM768Key)) {
+                features.add("session-key-mlkem");
+            }
+            return features;
+        }
 
         @CommandLine.Parameters(index = "0", arity = "0..1", description = "Feature to check (e.g., dpop). Omit to list all supported features.")
         private String feature;
@@ -87,12 +100,13 @@ class Command {
 
         @Override
         public Integer call() {
+            List<String> features = features();
             if (feature == null) {
-                printFeatures(FEATURES);
+                printFeatures(features);
                 return 0;
             }
 
-            Optional<String> canonical = FEATURES.stream().filter(f -> f.equalsIgnoreCase(feature)).findFirst();
+            Optional<String> canonical = features.stream().filter(f -> f.equalsIgnoreCase(feature)).findFirst();
             if (json) {
                 Map<String, Boolean> result = new LinkedHashMap<>();
                 // Emit the canonical feature name for recognized features so casing is stable
--- a/cmdline/src/test/java/io/opentdf/platform/CommandTest.java
+++ b/cmdline/src/test/java/io/opentdf/platform/CommandTest.java
@@ -95,7 +95,7 @@ class CommandTest {
         String out = captureStdout(() -> code[0] = new CommandLine(new Command()).execute("supports"));
 
         assertThat(code[0]).isEqualTo(0);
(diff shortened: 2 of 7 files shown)


Arpan0995:
I came across this PR while looking at ML-KEM adoption across Java libraries and read the rewrap path against FIPS 203 and the envelope format. The core choices look right: the ML-KEM session key is a fresh ephemeral per `unwrap()` call (matching the EC path's per-call `ECKeyPair`, which is the property you want for a rewrap session key); the envelope parsing in `HybridCrypto` is strict (canonical DER lengths, trailing bytes rejected, bounds-checked reads); `KemProviders.get()` fails with a clear `SDKException` when the `sdk-pqc-bc` module is absent rather than an NPE; and the new round-trip test covers all five key types. Using the FIPS 203 shared secret directly as the wrap key for pure ML-KEM is sound (it is a uniform 32-byte key by construction) and the ADR reference makes the intent clear.

One real issue and two smaller notes.

**1. The new double-checked locking leaves a read-side race on `clientPublicKey` (low, fail-closed).** In `generateSessionKeyMaterial` the initializing thread writes `decryptor` first and `clientPublicKey` second (`KASClient.java:157-158`), but the fast path reads `decryptor` at `:153` and, when it is non-null, returns `clientPublicKey` at `:162` without entering the lock. A thread that arrives between the two writes sees `decryptor != null`, skips the synchronized block, and returns a `SessionKeyMaterial` whose `publicKeyPem` is still null, so the rewrap request is built with `body.clientPublicKey = null` (`:202`). This is plain interleaving, not a memory-model corner case, and the window is wider than one instruction: it spans the whole Base64 PEM encoding on the right-hand side of `:158`. I checked it with a two-thread harness mirroring the exact field and write structure, forcing a reader into that window: it read the null public key in 19,677 of 20,000 forced races, and with the two writes swapped (so the volatile publish of `decryptor` covers the PEM) it read null in 0 of 20,000. The forced rate is not the production frequency, which depends on thread timing, but it isolates the write-order as the cause and confirms the swap removes it. The window only exists around first use of the RSA path and the result is a rejected request rather than anything unsafe, so this is low severity, but it is the exact class of mixed-state send the comment above the block says the locking is there to prevent. The minimal fix is to swap the two writes (writing `clientPublicKey` before `decryptor` makes the `decryptor` volatile read at `:153` guarantee visibility of the PEM); returning the PEM from inside the synchronized block, or storing both fields in a single volatile holder object, also works and is harder to regress.

**2. The SPI freezes the private key as an immutable `String` (design note, worth deciding now).** `KemProvider.KeyPairPem` (`KemProvider.java:63-70`) carries the ephemeral ML-KEM private key as a PEM `String`. Strings cannot be zeroized, so the decapsulation key lives on the heap until GC with no way to wipe it after the unwrap completes. The EC path has the same property today via key objects, so this is not a regression, but since this PR is the moment `KeyPairPem` enters the SPI contract, it is the cheapest time to choose `byte[]`/`char[]` (with a wipe after use) over `String` if you ever want that option; changing it later means breaking the interface.

**3. Negative tests (small).** `HybridCryptoGenerateKeyPairTest` round-trips the happy path. Given the parser already has strict failure modes, two cheap additions would pin the security-relevant behavior: unwrap with a different key pair of the same type must fail, and an envelope with a flipped byte in the ciphertext must fail rather than return a DEK. Both are a few lines in the same parameterized style.

The strict envelope parser and the per-call ephemeral are the parts that are easy to get wrong in a rewrap flow, and both look right here.

PR #391, opened 2026-08-20
feat(sdk): cache ListKeyAccessServers response during decryption

### LINKED ISSUE #390: feat(sdk): cache ListKeyAccessServers response during decryption
## Problem

Every call to `loadTDF()` triggers a `ListKeyAccessServers` RPC to the platform service to build a KAS allowlist. This happens in:

- `TDF.loadTDF()` in `sdk/src/main/java/io/opentdf/platform/sdk/TDF.java:546`

The result is never cached. There is no memoization and no deduplication of concurrent requests. If a caller decrypts 100 files, 100 identical RPCs are made to fetch the same KAS server list.

For comparison, KAS public keys *are* cached via `KASKeyCache` (`sdk/src/main/java/io/opentdf/platform/sdk/KASKeyCache.java`). The server list has no equivalent caching.

Callers can work around this by passing `WithKasAllowlist()` directly, but the default path always makes the network call.

## Proposed Solution

Add a cache for the `listKeyAccessServers()` result. The cache should:

- Be keyed by `platformUrl`
- Deduplicate concurrent in-flight requests for the same platform URL
- Be scoped to the `SDK` instance so it does not leak across unrelated sessions
- Be invalidatable for callers that need a fresh list

This would reduce redundant network calls without sacrificing security, since the KAS registry changes infrequently relative to decrypt operations.

See also: https://github.com/opentdf/web-sdk/issues/998 (same issue in the web SDK) and https://github.com/opentdf/platform/issues/3897 (same issue in the Go SDK).

eugenioenko:
## Motivation

Every `loadTDF` call triggers a `ListKeyAccessServers` RPC to build the KAS allowlist. The result is never cached, so decrypting 100 files makes 100 identical RPCs. KAS public keys are already cached via `KASKeyCache`; the allowlist had no equivalent.

Resolves #390. See also: https://github.com/opentdf/web-sdk/issues/998, https://github.com/opentdf/platform/issues/3897.

## PR Changes

- Add `KASAllowlistCache` class mirroring `KASKeyCache` (keyed by platform URL, 5-minute TTL)
- Add private `resolveKasAllowlist()` method on `SDK` that encapsulates cache check, RPC fallback, and cache store
- `SDK.loadTDF()` calls `resolveKasAllowlist()` to populate `config.kasAllowlist` before delegating to `TDF`
- No changes to `TDF`, `Services`, or existing tests
- Add `KASAllowlistCacheTest` with tests for basic store/get, expiration, clear, and multiple entries

## Summary by CodeRabbit

* **New Features**
  * Added automatic KAS allowlist resolution when loading TDFs.
  * Reuses recently retrieved allowlists for improved performance.
  * Refreshes cached allowlists after five minutes and supports clearing cached entries.
  * Reports clear errors when allowlist lookup fails.

### DIFF
--- a/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
@@ -11,6 +11,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
+// Mirrors the KASKeyCache pattern: null returns, LocalDateTime.now(), and 5-minute TTL.
 class KASAllowlistCache {
     private static final Logger log = LoggerFactory.getLogger(KASAllowlistCache.class);
     Map<String, TimeStampedAllowList> cache;
--- a/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
@@ -6,22 +6,18 @@ import org.slf4j.LoggerFactory;
 import java.time.LocalDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 // Mirrors the KASKeyCache pattern: null returns, LocalDateTime.now(), and 5-minute TTL.
 class KASAllowlistCache {
     private static final Logger log = LoggerFactory.getLogger(KASAllowlistCache.class);
-    Map<String, TimeStampedAllowList> cache;
-
-    public KASAllowlistCache() {
-        this.cache = new HashMap<>();
-    }
+    final Map<String, TimeStampedAllowList> cache = new ConcurrentHashMap<>();
 
     public void clear() {
-        this.cache = new HashMap<>();
+        cache.clear();
     }
 
     public Set<String> get(String platformURL) {
--- a/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
@@ -1,0 +1,58 @@
+package io.opentdf.platform.sdk;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+class KASAllowlistCache {
+    private static final Logger log = LoggerFactory.getLogger(KASAllowlistCache.class);
+    Map<String, TimeStampedAllowList> cache;
+
+    public KASAllowlistCache() {
+        this.cache = new HashMap<>();
+    }
+
+    public void clear() {
+        this.cache = new HashMap<>();
+    }
+
+    public Set<String> get(String platformURL) {
+        log.debug("retrieving allowlist for platformURL = [{}]", platformURL);
+        TimeStampedAllowList cachedValue = cache.get(platformURL);
+        if (cachedValue == null) {
+            log.debug("didn't find allowlist for platformURL = [{}]", platformURL);
+            return null;
+        }
+
+        LocalDateTime fiveMinAgo = LocalDateTime.now().minus(5, ChronoUnit.MINUTES);
+        if (fiveMinAgo.isAfter(cachedValue.timestamp)) {
+            log.debug("cached allowlist is too old timestamp = [{}] for platformURL = [{}]",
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
```diff
@@ -0,0 +1,58 @@
...
+            return null;
+        }
+
+        log.debug("successfully returned allowlist for platformURL = [{}]", platformURL);
+        return cachedValue.allowlist;
+    }
+
+    public void store(String platformURL, Set<String> allowlist) {
+        log.debug("storing allowlist into the cache for platformURL = [{}]", platformURL);
+        cache.put(platformURL, new TimeStampedAllowList(allowlist, LocalDateTime.now()));
```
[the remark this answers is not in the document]
eugenioenko: Good catch — fixed in 5b27531. The cache now stores an unmodifiable defensive copy, and `get()` returns a new `HashSet` each time.

sdk/src/main/java/io/opentdf/platform/sdk/SDK.java
```diff
@@ -136,10 +141,42 @@ public Optional<SimpleKasKey> getBaseKey() {
     }
 
     public TDF.Reader loadTDF(SeekableByteChannel channel, Config.TDFReaderConfig config) throws SDKException, IOException {
+        resolveKasAllowlist(config);
         var tdf = new TDF(services);
         return tdf.loadTDF(channel, config, platformUrl);
```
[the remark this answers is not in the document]
eugenioenko: This is pre-existing behavior — the original `TDF.loadTDF()` already calls `listKeyAccessServersBlocking` before reading the manifest (TDF.java:546). This PR doesn't change that ordering, it just caches the result. Any fix for plaintext TDF handling should be a separate issue.

sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
```diff
@@ -0,0 +1,62 @@
...
+class KASAllowlistCache {
+    private static final Logger log = LoggerFactory.getLogger(KASAllowlistCache.class);
+    Map<String, TimeStampedAllowList> cache;
+
+    public KASAllowlistCache() {
+        this.cache = new HashMap<>();
+    }
+
+    public void clear() {
+        this.cache = new HashMap<>();
```
pflynn-virtru: `cache` is a plain `HashMap`, but `KASAllowlistCache` is held as a field on `SDK` (`SDK.java:64`), so a single instance is shared for the lifetime of the client. Three things make that unsafe under concurrent `loadTDF` calls:

- `get()` **writes** to the map — the expiry path calls `cache.remove(platformURL)` on line 39 — so this isn't a read-mostly accessor that could get away without synchronization.
- `store()` writes on every miss.
- The field is non-`final` and `clear()` reassigns it rather than clearing in place, so a `clear()` on one thread may never become visible to another (no happens-before edge).

This matters specifically because of the PR's own motivation — "decrypting 100 files makes 100 identical RPCs". The natural way a caller speeds that up is a thread pool over one `SDK` instance, which is exactly the pattern that puts concurrent `get`/`store` on an unsynchronized `HashMap`: lost entries, or corruption during a resize.

Cheap fix:

```java
private final Map<String, TimeStampedAllowList> cache = new ConcurrentHashMap<>();

public void clear() {
    cache.clear();
}
```

I see the header comment notes this mirrors `KASKeyCache`, which has the same shape — so this may be deliberate consistency. Worth calling out that the exposure is different though: `KASKeyCache` is reached far less often than one lookup per `loadTDF`.

sdk/src/main/java/io/opentdf/platform/sdk/SDK.java
```diff
@@ -136,10 +141,42 @@ public Optional<SimpleKasKey> getBaseKey() {
...
+            config.kasAllowlist = cached;
+            return;
+        }
+
+        var request = ListKeyAccessServersRequest.newBuilder().build();
+        ListKeyAccessServersResponse response;
+        try {
+            response = RequestHelper.getOrThrow(
+                    services.kasRegistry().listKeyAccessServersBlocking(request, Collections.emptyMap()).execute());
+        } catch (Exception e) {
```
pflynn-virtru: narrow scope if you can

sdk/src/main/java/io/opentdf/platform/sdk/KASAllowlistCache.java
eugenioenko: Fixed in b5254ca — switched to `ConcurrentHashMap` with `final` field and `cache.clear()` instead of reassignment.

sdk/src/main/java/io/opentdf/platform/sdk/SDK.java
eugenioenko: Fixed in b5254ca — narrowed to `ConnectException`, matching the original `TDF.loadTDF` catch.

PR #393, opened 2026-08-27
feat(sdk): DSPX-4495 remove 64GiB TDF limit, use counter-based payload IVs (DSP…

sujankota:
Removes MAX_TDF_INPUT_SIZE (68719476736) and the size check in createTDF. That constant was GCM's per-invocation plaintext limit (2^39-256 bits) misapplied to the whole TDF input; each segment is its own invocation and is capped at 4MiB by Config.MAX_SEGMENT_SIZE, so the bound was never relevant. SDK.DataSizeNotSupported is retained as public API but is no longer thrown.

Replaces the random per-segment AES-GCM nonce with a deterministic unsigned 96-bit big-endian counter (TDF.IvCounter). NIST SP 800-38D prefers the deterministic construction; the RBG-based one it replaces carries a birthday bound that a counter does not have.

Metadata is encrypted with the per-split symKey while the payload uses the XOR of all split keys, so with a single key split the two are the same key. IV 0 is therefore reserved for the metadata and payload segments start at IV 1.

The counter enforces the SP 800-38D 8.3 cap of 2^32 invocations per key and refuses to wrap, so an IV can never be issued twice. Neither limit is reachable in practice - at the 16KiB minimum segment size the cap is 64TiB of input - but the invariant now holds by construction rather than by assumption.

Every segment is still prefixed with its 12-byte IV, so the wire format is unchanged and existing TDFs continue to decrypt.

## Summary by CodeRabbit

* **New Features**
  * TDF creation supports inputs spanning many segments.
  * Added reliable ZIP64 support for reading and writing large archives.
  * Encryption uses sequential IV allocation with safeguards against reuse.

* **Bug Fixes**
  * Improved handling of large ZIP offsets, sizes, and entry metadata.
  * Added stricter AES-GCM validation for IV and authentication tag sizes.

* **Documentation**
  * Clarified that the data-size exception is deprecated, never thrown, and retained for legacy compatibility.

### DIFF
--- a/sdk/src/main/java/io/opentdf/platform/sdk/ZipReader.java
+++ b/sdk/src/main/java/io/opentdf/platform/sdk/ZipReader.java
@@ -52,6 +52,15 @@ public class ZipReader {
         return result.intValue();
     }
 
+    /**
+     * Reads a 32-bit zip field as the unsigned value it is on the wire. {@link #readInt()}
+     * sign-extends, which silently turns offsets and sizes at or above 2 GiB into negative
+     * numbers.
+     */
+    private long readUnsignedInt() throws IOException {
+        return readInt() & 0xFFFFFFFFL;
+    }
+
     final ByteBuffer shortBuf = ByteBuffer.allocate(Short.BYTES).order(ByteOrder.LITTLE_ENDIAN);
 
     private short readShort() throws IOException {
@@ -61,6 +70,14 @@ public class ZipReader {
         }
         shortBuf.flip();
         return shortBuf.getShort();
+    }
+
+    /**
+     * Reads a 16-bit zip field as the unsigned value it is on the wire. See
+     * {@link #readUnsignedInt()}.
+     */
+    private int readUnsignedShort() throws IOException {
+        return readShort() & 0xFFFF;
     }
 
     private static class CentralDirectoryRecord {
@@ -79,7 +96,10 @@ public class ZipReader {
     private static final int CENTRAL_FILE_HEADER_SIGNATURE =  0x02014b50;
 
     private static final int LOCAL_FILE_HEADER_SIGNATURE =  0x04034b50;
-    private static final int ZIP64_MAGICVAL = 0xFFFFFFFF;
+    /** Sentinel written into a 32-bit field whose real value lives in the zip64 extra field. */
+    private static final long ZIP64_MAGICVAL = 0xFFFFFFFFL;
+    /** The same sentinel for a 16-bit field, which is only two bytes wide. */
+    private static final int ZIP64_MAGIC_SHORT = 0xFFFF;
     private static final int ZIP64_EXTID= 0x0001;
 
     CentralDirectoryRecord readEndOfCentralDirectory() throws IOException {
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/io/opentdf/platform/sdk/ZipReader.java
```diff
@@ -177,12 +220,16 @@ public InputStream getData() throws IOException {
                     + Short.BYTES
                     + Integer.BYTES);
 
-            long compressedSize = readInt();
-            long uncompressedSize = readInt();
-            int filenameLength = readShort();
-            int extrafieldLength = readShort();
+            long compressedSize = readUnsignedInt();
```
dmihalcik-virtru, line 223: These useless assignments keep with the existing pattern of labelling the skipped zip fields.

sdk/src/main/java/io/opentdf/platform/sdk/ZipReader.java
```diff
@@ -177,12 +220,16 @@ public InputStream getData() throws IOException {
...
+            int extrafieldLength = readUnsignedShort();
+
+            return zipChannel.position() + filenameLength + extrafieldLength;
+        }
 
-            final long startPosition = zipChannel.position() + filenameLength + extrafieldLength;
+        public InputStream getData() throws IOException {
+            final long startPosition = findStartOfData();
             final long endPosition = startPosition + fileSize;
             final ByteBuffer buf = ByteBuffer.allocate(1);
```
pflynn-virtru, line 234: A `buf.clear();` before the read loop below fixes the single-byte `read()` — `buf` is hoisted out here, so after the first call `hasRemaining()` is false and every subsequent `read()` returns the same byte without touching the channel.

pflynn-virtru (dismissed):

pflynn-virtru:
(AI generated below, unvetted)
Removing `MAX_TDF_INPUT_SIZE` leaves nothing bounding the manifest, which is assembled in memory. A 64 GiB input at `Config.MIN_SEGMENT_SIZE` (16 KiB) produces ~4.19M `Manifest.Segment` objects plus a `gson.toJson` string of roughly 90 chars each — well over 2 GB across the segment list, the JSON `String`, and its UTF-8 byte copy. On a default heap that dies with `OutOfMemoryError` at `gson.toJson(...)` *after* the entire payload has already been streamed to `outputStream`. Previously the same input threw `DataSizeNotSupported` before writing anything.

Two suggestions, one cheap and one optional.

**1. Bound the segment count.** The input is an `InputStream`, so there's no size to pre-check — but you can fail deterministically instead of via `OutOfMemoryError`. A constant next to `MAX_GCM_INVOCATIONS_PER_KEY`:

```java
/**
 * The manifest holds one record per segment and is assembled in memory, so the segment count
 * is bounded well below {@link #MAX_GCM_INVOCATIONS_PER_KEY}. At the default segment size this
 * is 4 TiB of payload; a caller who hits it should raise the segment size rather than lower it.
 */
static final int MAX_SEGMENTS = 1 << 21;
```

and a check at the `segments.add(segmentInfo)` in `createTDF`:

```java
var segments = tdfObject.manifest.encryptionInformation.integrityInformation.segments;
if (segments.size() >= MAX_SEGMENTS) {
    throw new SDKException("exceeded the maximum of " + MAX_SEGMENTS
            + " segments; increase the segment size for payloads this large");
}
segments.add(segmentInfo);
```

`SDKException` rather than reviving `DataSizeNotSupported` — it matches how `IvCounter` reports its bound, and keeps the story clean: no fixed input-size limit, but there is a segment-count limit. `SDK.java` would need a word change, since that bound stops being "practical rather than fixed."

Sizing: `2^21` covers 4 TiB at the 2 MiB default and 8 TiB at `MAX_SEGMENT_SIZE`. It does reject 64 GiB at `MIN_SEGMENT_SIZE` (4.19M segments) — deliberately, since that's the config that OOMs today, and a clear exception pointing at the segment size beats dying at `gson.toJson` after the whole payload is on the wire.

**2. Stream the manifest (optional, later).** The manifest is written after the payload, and `ZipWriter` already has a `stream()` path, so a gson `JsonWriter` writing straight into the entry would drop the intermediate `String` and its UTF-8 `byte[]` — roughly a 3x cut in peak memory, leaving only the segment list. Bigger change, and I haven't checked whether anything downstream needs the serialized manifest bytes, so the ceiling is the one worth landing first.

dmihalcik-virtru:
yeah, I filed https://virtru.atlassian.net/browse/DSPX-4584 for the streaming issue - I suspect Java isn't the only SDK which will suffer from this or a similar issue.

pflynn-virtru (approved):

PR #399, opened 2026-09-09
feat(sdk): DSPX-4584 stream the TDF manifest, remove the ~21.7M segment ceiling

sujankota:
The manifest was assembled entirely in memory on both paths, so a payload past ~21.7M segments (41 TiB at 2 MiB segments) hard-failed with OutOfMemoryError: Requested array size exceeds VM limit.

- serialize the manifest straight into the zip entry and parse it from a Reader, so neither side materializes it as a Java String
- add Manifest.Segments, a compact fixed-stride AbstractList<Segment> backed by packed hash bytes (~24 B/segment instead of ~176), with an ArrayList fallback for manifests this SDK did not write
- materialize the aggregate hash once and share it with the root signature and every assertion instead of copying it per assertion

Manifest JSON bytes are unchanged. The 0.manifest.json zip entry is now a streamed data-descriptor entry, the same form 0.payload already used.

## Summary by CodeRabbit

- **New Features**
  - Large manifests can now be read and written as streams, improving support for very large TDF files.
  - Consistent manifest segments use less memory while preserving the same serialized format.
  - Segment hashes are aggregated consistently during encryption and decryption.

- **Breaking Changes**
  - Manifest access now uses reader and output-stream APIs instead of complete strings; callers must close manifest streams.
  - Unsupported root or segment integrity algorithms are now rejected explicitly.

- **Compatibility**
  - The deprecated string-based manifest append method is available again for existing integrations.


sdk/src/main/java/io/opentdf/platform/sdk/TDFReader.java
```diff
@@ -62,8 +69,10 @@ int readPayloadBytes(byte[] buf) {
     }
 
     PolicyObject readPolicyObject() {
-        String manifestJson = manifest();
-        Manifest manifest = Manifest.readManifest(manifestJson);
-        return Manifest.decodePolicyObject(manifest);
+        try (Reader manifestJson = manifest()) {
+            return Manifest.decodePolicyObject(Manifest.readManifest(manifestJson));
```
dmihalcik-virtru, line 73:
```suggestion
        try (Reader manifestRaw = manifest()) {
            return Manifest.decodePolicyObject(Manifest.readManifest(manifestRaw));
```

sdk/src/main/java/io/opentdf/platform/sdk/TDF.java
```diff
@@ -472,6 +481,23 @@ public PolicyObject readPolicyObject() {
         }
     }
 
+    /**
+     * Concatenates every segment hash, which is what the root signature and the
+     * assertion signatures are computed over.
+     */
+    private static byte[] aggregateSegmentHashes(List<Manifest.Segment> segments, boolean isEncrypted) {
```
dmihalcik-virtru, line 488: Instead of creating this in memory, you can just call update multiple times on the `Mac` object.

PR #400, opened 2026-09-10
feat(sdk,cmdline): settable segment and root integrity algorithms (DSPX-4736)

dmihalcik-virtru:
## Summary

Adds explicit controls for the two integrity algorithms a ZTDF writer picks, and makes an unsupported choice fail loudly instead of being silently accepted.

## Rationale

The two algorithms are not interchangeable, and the API previously did not say so:

| | allowed | why |
|---|---|---|
| **segment** | `HS256`, `GMAC` | a segment hash is computed over ciphertext AES-GCM actually produced, so "GMAC" means reading back a real authentication tag |
| **root** | `HS256` only | the root signature covers the aggregate of the segment hashes, which never passes through AES-GCM — there is no tag to recover, so GMAC would authenticate nothing |

**`sdk`**
* `Config.withSegmentIntegrityAlgorithm(HS256|GMAC)`
* `Config.withRootIntegrityAlgorithm(HS256)` — throws `IllegalArgumentException` for anything else. The option exists so callers can state the choice explicitly and a CLI can surface the refusal, not to widen it. It is self-contained and does not reach into `TDF`.

**`cmdline`**
* `encrypt --root-integrity-algorithm` and `--segment-integrity-algorithm`, with a case-insensitive `ITypeConverter` so the flags behave like the equivalents in the other OpenTDF CLIs, and `${COMPLETION-CANDIDATES}` so the accepted values render in help.
* The root algorithm is validated **before** `buildSDK()`, so an unsupported value is a picocli usage error (exit 2) rather than a failure after a platform round trip.
* `encryptHelp_advertisesIntegrityAlgorithmFlags` pins the literal flag names into the rendered encrypt help, because the cross-SDK xtest feature detectors find them by grepping it. Merging this flips the `integrity_algs` / `gmac_root_rejected` gates on and stops those cells silently skipping.

## Scope

This is the **control surface** for evaluating DSPX-4703, deliberately separated from the fix. It changes what a writer may *choose*; it does not change how a manifest that already declares a GMAC root is verified on **read**. That is the stacked follow-up.

Splitting it this way means the xtest exploit cases can run against this branch for a live "before" baseline, then against the fix branch to watch them go red → green.

## Public API notes

Two new static factory methods on `Config`. Nothing existing changes shape or default behaviour: the segment default stays `GMAC`, the root default stays `HS256`.

## Test plan

```bash
mvn -q compiler:compile compiler:testCompile -pl sdk,cmdline
mvn -q surefire:test -pl cmdline -Dtest=CommandTest
```

`CommandTest`: 26 tests, 0 failures — run with this commit as `HEAD` and the DSPX-4703 fix commit entirely absent, to confirm the branch stands alone.

Coverage added: GMAC rejected for root in three casings, HS256 accepted in two casings (and reaching the *later* missing-credentials error, proving it was not rejected), both algorithms accepted for segment, an unknown algorithm rejected, converter case-insensitivity, and the help-rendering assertion.

## Summary by CodeRabbit

* **New Features**
  * Added command-line options to configure root and segment integrity algorithms during encryption.
  * Integrity algorithm values now accept mixed casing and surrounding whitespace.
  * Segment integrity supports HS256 and GMAC, while root integrity supports HS256.

* **Bug Fixes**
  * Invalid or unknown integrity algorithm values are rejected with clear command-line errors.
  * Encryption configuration validation now occurs before processing begins.

Cross-SDK coverage lives in opentdf/tests#594.


dmihalcik-virtru:
these should be fixed values for now

